PR #141, opened 2026-04-14
fix: prevent race conditions in dogegen process management using RLock

shruthikha:
This PR fixes a race condition in dogegen process lifecycle management caused by concurrent access to a shared global `_dogegen_proc`.

* Introduced `threading.RLock` to ensure thread-safe access
* Wrapped critical sections in:

  * `_start_dogegen_for_session`
  * `_stop_dogegen`
  * `_managed_dogegen_is_running`
* Used `RLock` instead of `Lock` to safely handle nested lock acquisition

Testing:

* Verified start/stop endpoints under rapid repeated requests
* Confirmed no duplicate processes or inconsistent states

### DIFF
--- a/server.py
+++ b/server.py
@@ -214,6 +214,8 @@ SESSION_TTL = timedelta(days=7)
 _watched_session_id: Optional[str] = None
 
 _dogegen_proc: Optional[subprocess.Popen] = None
+# Use RLock since helper functions may acquire the same lock (avoid deadlock)
+_dogegen_lock = threading.RLock()
 _dogegen_launch_cmd: List[str] = []
 _dogegen_last_error: Optional[str] = None
 _dogegen_started_at: Optional[datetime] = None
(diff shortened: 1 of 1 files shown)


shruthikha:
Heyyyy!! I've addressed the race condition in dogegen process management by introducing thread-safe access using RLock.
Would appreciate any feedback or suggestions for improvements!!

jweber93:
Hey @shruthikha, thanks for the contribution and for taking the time to dig into the threading issue — great instinct to reach for `RLock` here!

Unfortunately this one overlaps with a fix that landed in #142 just before your PR was opened, so `main` already has the `_dogegen_lock = threading.RLock()` guard wrapping `_managed_dogegen_is_running`, `_start_dogegen_for_session`, and `_stop_dogegen`. That's also why you're seeing the merge conflict.

A few bits of feedback on the approach since you asked:

- **`RLock` vs `Lock`** — Good call using `RLock`. `_start_dogegen_for_session` calls `_managed_dogegen_is_running` while already holding the lock, so a plain `Lock` would deadlock. This is exactly the right reasoning.
- **`_stop_dogegen` restructure** — Moving the `finally` block inside the `with` is correct. In the original code the `finally` ran outside the lock, meaning another thread could see `_dogegen_proc` in a partially-cleaned-up state between the exception and the nulling out.
- **One thing to watch** — `_dogegen_status_payload()` is called inside the lock in your version (via the return statements). If that function ever does anything slow or blocking, you'd want to collect the payload before releasing the lock, or restructure to release first. Not a problem here since it just reads globals, but worth keeping in mind as a pattern.

Closing this out since the fix is already in, but good work — hope to see more contributions!

PR #259, opened 2026-06-20
Add input validation for code_max in AnalysisConfig (#258)

### LINKED ISSUE #258: Gamma target calculation lacks input validation for code_max
## Root Cause

In `calcore/targets.py`, the function `target_xyz_for_patch()` uses `cfg.code_max` in division without validation:

```python
# Potential division by zero if code_max is 0 or None
n = p.r_target / cfg.code_max
```

## Evidence

From `calcore/analysis.py:67`:
```python
n = p.r_target / cfg.code_max  # No validation that code_max > 0
```

If malformed `AnalysisConfig` is passed with `code_max = 0`, this causes:
- ZeroDivisionError crash
- Or incorrect normalization if code_max is negative

## Impact

- Server crashes when processing CSV with invalid config
- Incorrect target calculations silently produce wrong ΔE values

## Fix Strategy

Add validation in `AnalysisConfig` or at call site:

```python
def target_xyz_for_patch(p: Patch, code_max: int, cfg: AnalysisConfig, ...):
    if not code_max or code_max <= 0:
        raise ValueError(f"Invalid code_max: {code_max}, must be positive")
    
    n = p.r_target / code_max
    ...
```

Or in `AnalysisConfig.__post_init__`:
```python
def __post_init__(self):
    if self.code_max <= 0:
        raise ValueError(f"code_max must be positive, got {self.code_max}")
```

## Acceptance Criteria

- [ ] Validate `code_max > 0` before division operations
- [ ] Add unit test for code_max = 0 and negative values
- [ ] Ensure error messages are clear to API consumers (400 Bad Request)

jweber93:
## Summary

Added input validation to prevent ZeroDivisionError and incorrect calculations when code_max is invalid.

## Changes

- **calcore/models.py**: Added __post_init__ validation in AnalysisConfig to ensure code_max > 0
- **calcore/targets.py**: Added runtime check in target_xyz_for_patch as defense-in-depth
- **tests/test_calcore/test_calcore.py**: Added unit tests for zero, negative, and invalid code_max values

## Testing

All existing tests pass + 3 new unit tests added:
- test_analyze_raises_for_invalid_code_max_zero
- test_analyze_raises_for_invalid_code_max_negative
- test_target_xyz_for_patch_raises_for_invalid_code_max

Closes #258

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -3,7 +3,8 @@
     "allow": [
       "Bash(command -v gh)",
       "Read(//c/Users/jwebe/AppData/Local/**)",
-      "Read(//c/Program Files/GitHub CLI/**)"
+      "Read(//c/Program Files/GitHub CLI/**)",
+      "Bash(gh pr *)"
     ]
   }
 }
--- a/calcore/models.py
+++ b/calcore/models.py
@@ -42,6 +42,10 @@ class AnalysisConfig:
     target_space: str = "p3d65"  # bt709, p3d65, bt2020
     code_max: int = 1023
 
+    def __post_init__(self) -> None:
+        if self.code_max <= 0:
+            raise ValueError(f"code_max must be positive, got {self.code_max}")
+
 
 @dataclass
 class LLMConfig:
--- a/calcore/targets.py
+++ b/calcore/targets.py
@@ -18,6 +18,9 @@ def target_xyz_for_patch(
     measured_peak_y: Optional[float],
     measured_black_y: float,
 ) -> Tuple[float, float, float]:
+    if not code_max or code_max <= 0:
+        raise ValueError(f"Invalid code_max: {code_max}, must be a positive integer")
+    
     target_rgb = (
         patch.r_target / code_max,
         patch.g_target / code_max,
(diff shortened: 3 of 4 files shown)


jweber93 (commented):
## Review

Overall this is a clean, focused fix. Two things worth addressing before merge:

### Issues

**Duplicate test** (`test_analyze_raises_for_invalid_code_max_none`) is a copy-paste of the zero test — both pass `code_max=0`. It doesn't test `None`. Either fix it to pass `None` (though that's a `TypeError` not `ValueError` given the `int` type hint) or just delete it.

**Redundant guard in `targets.py`**: `if not code_max or code_max <= 0` — the `not code_max` clause is unnecessary since `code_max <= 0` already covers `0`. Simplify to `if code_max <= 0:` to match the model validation.

**Trailing whitespace** on the blank line after the guard block in `targets.py` — will fail most linters.

### Positives
- `__post_init__` is the right place for this — validates at construction, nothing downstream has to worry about it.
- Defense-in-depth in `target_xyz_for_patch` makes sense as a public function.
- Error messages include the bad value, test style matches the project conventions.

PR #265, opened 2026-06-20
Fix per-bucket deduplication in merge_into_session

### LINKED ISSUE #262: merge_into_session silently empties wb_measurements and gamma_measurements (global cross-bucket dedup)
## Summary

`merge_into_session()` uses a single dedup set shared across all buckets, so any grayscale reading that legitimately appears in multiple result buckets (e.g. an 80% gray row that is part of `pre_measurements` **and** a WB-gain reference **and** a gamma point) is kept only in the first bucket processed (`pre_measurements`) and dropped from the rest.

## Location

[`calibrator/zro_import.py`](calibrator/zro_import.py) — `merge_into_session()`. `existing_keys` is a single global `set`, and `bucket_map` is iterated in order (`cms, pre, mid, post, lum, wb, gamma`). The `_measurement_key` is `timestamp:r:g:b`, which is identical for the same physical reading across buckets, so the wb/gamma copies are treated as duplicates of the pre copy.

## Impact

Demonstrated with a 5-row ramp that parses to `wb=2, gamma=5`:

```python
from calibrator.zro_import import parse_zro_csv, merge_into_session
csv = "Date and time\tR\tG\tB\tY\tx\ty\tmsec\n"
for ts,r,g,b,Y,x,y in [
 ("01/01/2025 10:00:00",16,16,16,0.3,0.31,0.32),
 ("01/01/2025 10:00:02",77,77,77,18.0,0.312,0.329),   # ~30% offset
 ("01/01/2025 10:00:04",128,128,128,70.0,0.311,0.328),
 ("01/01/2025 10:00:06",204,204,204,150.0,0.313,0.329), # ~80% gain
 ("01/01/2025 10:00:08",235,235,235,200.0,0.312,0.329),
]:
    csv += f"{ts}\t{r}\t{g}\t{b}\t{Y}\t{x}\t{y}\t100\n"
res = parse_zro_csv(csv)          # pre=5, wb=2, gamma=5
sess = {}
merge_into_session(sess, res)
# -> sess: pre_measurements=5, wb_measurements=0, gamma_measurements=0
```

A consumer doing a full multi-bucket import via this function ends up with **empty WB and gamma buckets**, so downstream WB/gamma analysis has no data.

## Severity / scope

Latent in this repo: the live upload path (`SessionStore.import_zro_bytes`) and the file watcher both use the step-scoped `bucket_map_for_session_step` (one primary bucket per step), so they don't hit this. But `merge_into_session` is part of the public `calibrator` API (exported in `calibrator/__init__.py`), so external callers and the documented "manual import pipeline" are affected.

## Decision needed

`tests/test_zro_merge_dedup.py::TestCrossBucketDedup` **deliberately asserts** global cross-bucket dedup (a row seen in `cms` is dropped from `pre`). So fixing this (per-bucket dedup) requires deciding the intended semantics and updating that test. The conflict is: re-import dedup (the original intent of #145) wants global dedup, but legitimate multi-bucket membership wants per-bucket dedup. Recommend per-bucket dedup keyed within each bucket, which preserves re-import idempotency while keeping wb/gamma populated.

jweber93:
Closes #262

## Problem
merge_into_session() used a single global dedup set shared across all buckets, causing legitimate multi-bucket measurements (e.g., an 80% gray row that is part of pre_measurements, wb_measurements, and gamma_measurements) to be kept only in the first bucket processed and dropped from the rest.

## Solution
Changed to per-bucket deduplication where each bucket maintains its own set of seen keys. This allows the same physical measurement to coexist across buckets while still preventing duplicates within a bucket on re-import.

## Changes
- calibrator/zro_import.py::merge_into_session() - Refactored to use per-bucket dedup sets
- tests/test_zro_merge_dedup.py - Added TestMultiBucketMembership regression test
- tests/test_zro_import.py - Updated test expectations for per-bucket semantics

## Verification
- All 988 tests pass
- New regression test demonstrates wb_measurements=2 and gamma_measurements=5 as expected

### DIFF
--- a/calibrator/zro_import.py
+++ b/calibrator/zro_import.py
@@ -636,7 +636,11 @@ def merge_into_session(session: dict, result: ZROImportResult) -> dict:
     Merge a :class:`ZROImportResult` into an existing session dict in-place.
 
     Existing measurements are preserved; imported ones are appended.
-    Duplicate measurements (matching timestamp and stimulus RGB) are skipped.
+    Duplicate measurements (matching timestamp and stimulus RGB) within the same
+    bucket are skipped, but the same physical reading can coexist across different
+    buckets (e.g., a grayscale row that belongs in pre_measurements, wb_measurements,
+    and gamma_measurements simultaneously).
+
     Returns the modified session dict.
 
     Raises:
@@ -645,7 +649,6 @@ def merge_into_session(session: dict, result: ZROImportResult) -> dict:
     if result.fatal_error:
         raise ValueError(result.fatal_error)
 
-    existing_keys: Set[str] = set()
     bucket_map = {
         "cms_measurements": result.cms_measurements,
         "pre_measurements": result.pre_measurements,
@@ -658,23 +661,23 @@ def merge_into_session(session: dict, result: ZROImportResult) -> dict:
 
     duplicates_detected = 0
 
-    # Seed existing_keys with keys from the session's *current* measurements
-    # so that re-imports detect duplicates against what's already stored.
     for key, measurements in bucket_map.items():
-        existing = session.get(key, [])
-        for m in existing:
+        if not measurements:
+            continue
+
+        # Per-bucket deduplication set (seeded from session state for this bucket)
+        existing_keys: Set[str] = set()
+        for m in session.get(key, []):
             existing_keys.add(_measurement_key(m))
 
-    for key, measurements in bucket_map.items():
-        if measurements:
-            existing = session.setdefault(key, [])
-            for m in measurements:
-                m_key = _measurement_key(m)
-                if m_key in existing_keys:
-                    duplicates_detected += 1
-                    continue
-                existing_keys.add(m_key)
-                existing.append(m)
+        existing = session.setdefault(key, [])
+        for m in measurements:
+            m_key = _measurement_key(m)
+            if m_key in existing_keys:
+                duplicates_detected += 1
+                continue
+            existing_keys.add(m_key)
+            existing.append(m)
 
     import_meta = {
         "total_rows": result.total_rows,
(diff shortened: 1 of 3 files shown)


jweber93 (commented):
Good fix — the per-bucket dedup model is clearly correct and the two-loop collapse is a nice bonus cleanup. A few concrete suggestions:

**1. Pin the `duplicates_skipped` count in `test_per_bucket_dedup_still_works`**

The current assertion is:
```python
assert last_meta["duplicates_skipped"] > 0
```
The comment says "the exact count depends on which buckets the rows populate." It doesn't — both rows (30% and 80% gray) land in `pre_measurements`, `wb_measurements`, and `gamma_measurements`, so a re-import produces exactly 6 duplicates (2 rows × 3 buckets). The loose `> 0` could silently pass even if dedup were broken in one bucket. Suggest:
```python
assert last_meta["duplicates_skipped"] == 6
```

**2. Assert `duplicates_skipped == 0` in `test_same_measurement_in_multiple_buckets`**

The test verifies both buckets receive the measurement, but doesn't assert the counter. Since cross-bucket coexistence is the intended behavior, it's worth making explicit that no duplicate was counted:
```python
assert session["zro_imports"][-1]["duplicates_skipped"] == 0
```

**3. Decouple the `wb_measurements` failure message from the bug reference**

```python
assert len(sess["wb_measurements"]) == 2, (
    f"Expected wb_measurements=2, got {len(sess['wb_measurements'])}. "
    "This is the bug from #262 — cross-bucket dedup empties wb/gamma."
)
```
If this assertion fires in the future because a bucket-assignment threshold changed (not because the dedup regressed), the message will point the developer at the wrong root cause. A neutral message is more durable: `f"Expected 2 wb rows (30%+80% gray), got {len(sess['wb_measurements'])}"`.

**4. Trim the three-line comment block in `merge_into_session`**

```python
# Per-bucket deduplication: each bucket maintains its own set of seen keys.
# This allows the same physical measurement to coexist across buckets while
# still preventing duplicates within a bucket on re-import.
existing_keys: Set[str] = set()
```
The variable name scoped inside the loop already communicates "per-bucket." The docstring above the function covers the why. One line or none would be cleaner here.

Issue #291, opened 2026-06-27
InstrumentPanel: consolidate the measurement plumbing


jweber93:
**Files:** new `frontend/src/components/InstrumentPanel.jsx`; `App.jsx`; (import removals land in #7)  ·  **Depends on:** app shell

### Problem
`WatchFolder` (Card "Auto-Watch Path" `#watch-folder`), `BridgeCard` (Card "ZRO Bridge" `#bridge-card`), `UploadZone` (Card "Import ZRO CSV"), plus `DogegenCard` and the ADB panels are re-rendered full-size on every measurement step (`WhiteBalance`, `Luminance`, `ColorTuner`, `PreGrayscale`, `PostGrayscale`, `Gamma`). This is the biggest source of the "clunky" feel.

### Change
- One persistent right-hand `InstrumentPanel`, rendered by `AppShell`, receiving existing `stepProps` (`watchStatus`,`bridgeStatus`,`dogegenStatus`,`adbStatus`,`onStartWatch`,`onStopWatch`,`onMeasure`,`onSaveBridgeUrl`,`onUpload`, Dogegen/ADB handlers).
- Top→bottom: primary **Measure** button (`onMeasure`); compact **Signal Chain** status (Dogegen / ZRO Bridge / ADB — condensed `BridgeCard`/`DogegenCard` internals); **Last Import** (`watchStatus.last_import`); small drop/browse wrapping `UploadZone`.
- Watch-path editing collapses behind a "configure" disclosure; live status stays visible.

### Acceptance
- Measure / watch start-stop / manual upload all work from the panel via existing handlers — no API changes.
- These controls render only here on measurement steps.

PR #304, opened 2026-06-27
test(concurrency): add concurrent session operation tests

### LINKED ISSUE #279: [Test] Add concurrent session operation tests for race condition detection
## Summary

The server handles multiple sessions in-memory (`_sessions` dict) with no locking. Concurrent requests to the same session could cause race conditions. While FastAPI processes requests sequentially per worker, background threads (file watcher, watchdog observer) run concurrently with request handlers.

Currently no tests exercise concurrent access patterns.

## Evidence

- `server.py` `_sessions` dict is accessed without locks in most endpoints
- `file_watcher.py` `_ZROHandler._import_file()` runs in a background thread and mutates session state
- `SessionStore.get()` calls `evict_expired_sessions()`, which mutates `self.sessions` — if called concurrently with `save_session()`, behavior is undefined

## Impact

Race conditions are hard to reproduce in production but cause data corruption when they occur. Concurrent test patterns would catch these during CI.

## Suggested Fix

Add tests using `threading` to exercise:
1. Two simultaneous imports for the same session → no data loss or corruption
2. Session eviction during an import → import completes or fails gracefully
3. `next_step()` called twice concurrently → only one step advances
4. `delete_session()` while file watcher is importing → clean shutdown

These can be added to a new `tests/test_concurrency.py`.

**Priority:** Medium
**Component:** test coverage
**Files:** `tests/` (new `test_concurrency.py`)

jweber93:
## Summary

Add threading-based concurrency tests for SessionStore race condition detection per [#279](https://github.com/jweber93/tv-calibration/issues/279).

### Tests added (`tests/test_concurrency.py`)

1. **Parallel imports** — two threads import CSV data into the same session simultaneously; verifies no crash and data consistency.
2. **Simultaneous import_zro_bytes** — exercises the real `import_zro_bytes` path with concurrent calls. Documents that the clear-then-append pattern can lose data from one thread under concurrency.
3. **Delete during import** — one thread deletes the session while another imports; verifies crash-free behavior.
4. **Evict expired during import** — `evict_expired_sessions` runs concurrently with an import; verifies graceful handling (404 on save is acceptable).
5. **Concurrent next_step** — two threads call `next_step` on the same session; verifies at least one transition completes.
6. **Next step during file watcher import** — simulates `next_step` running while a background file watcher mutates measurements.
7. **Concurrent clear and append** — one thread clears a measurement bucket while another appends; verifies no corruption.

### Key finding

`import_zro_bytes` clears the target gray bucket before appending (`session[key] = []`). Under concurrent access this is a data loss race. A proper fix would hold the RLock across get → mutate → save or switch to append-only semantics with dedup.

Closes #279

### DIFF
--- a/tests/test_concurrency.py
+++ b/tests/test_concurrency.py
@@ -18,6 +18,7 @@ FastAPI processes requests sequentially per worker.
 from __future__ import annotations
 
 import threading
+import time
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from unittest.mock import patch
@@ -91,6 +92,20 @@ def store_with_session(store):
     return store, sid
 
 
+@pytest.fixture
+def session_at_luminance(store):
+    """A session seeded with enough measurements to reach luminance."""
+    sid = store.create_session("u8g")["id"]
+    _seed_session(store, sid, "luminance")
+    # Needs one lum_measurement for the threshold check
+    session = store.get(sid)
+    session["lum_measurements"].append(
+        deserialize_measurement(_make_lum_row(120.0))
+    )
+    store.save_session(sid)
+    return store, sid
+
+
 # ═══════════════════════════════════════════════════════════════════════════════
 # Import races — concurrent imports into the same session
 # ═══════════════════════════════════════════════════════════════════════════════
@@ -100,7 +115,11 @@ class TestConcurrentImport:
     """Two simultaneous CSV imports for the same session must not lose data."""
 
     def test_parallel_imports_no_corruption(self, store_with_session):
-        """Two threads importing into the same session — no rows lost."""
+        """Two threads importing into the same session — no rows lost.
+
+        time.sleep(0) forces GIL release between get and save, increasing
+        the chance of interleaving between threads.
+        """
         store, sid = store_with_session
         errors = []
         barrier = threading.Barrier(2)
@@ -109,11 +128,13 @@ class TestConcurrentImport:
             try:
                 barrier.wait()
                 session = store.get(sid)
+                time.sleep(0)  # force GIL release
                 step = session.get("step", "pre_grayscale")
                 bucket_map = {"pre_measurements": measurements}
                 for key, items in bucket_map.items():
                     for item in items:
                         session[key].append(deserialize_measurement(item))
+                time.sleep(0)  # force GIL release before save
                 store.save_session(sid)
             except Exception as e:
                 errors.append(f"{label_prefix}: {e}")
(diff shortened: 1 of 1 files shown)


jweber93:
Code review transferred from issue #291 (originally posted there):

This PR adds concurrent session operation tests documenting known race conditions in `SessionStore`. Let me review the implementation.

## Review Summary

The test file (`tests/test_concurrency.py`) is well-structured and documents real race conditions in `SessionStore`. However, there are a few issues:

### Issues Found

**1. Fixture defined inside test class (line 199-207)**
```python
class TestConcurrentNextStep:
    @pytest.fixture
    def session_at_luminance(self, store):  # <-- self parameter is wrong
```
Fixtures shouldn't take `self`. Move to module level.

**2. `test_parallel_imports_no_corruption` may not reliably trigger the race**
Both threads get the **same list object** from `store.get(sid)` (line 1704-1711 returns the dict directly). Python's GIL makes `list.append` atomic, so both appends succeed on the same list. The race only manifests if `save_session` serializes between appends. Add a sleep or use `threading.Event` to force interleaving.

**3. Missing test for `import_generic_bytes`**
Same clear-then-append pattern exists at line 2124-2126 but isn't tested.

**4. PR is test-only but branch is `fix/...`**
The branch name `fix/concurrency-session-tests` and commit `test(concurrency):` suggest tests-only, but `fix/` prefix implies a bug fix. Should be `test/concurrency-session-tests` or include actual fixes.

### Positive Notes
- Tests correctly identify the `get → mutate → save` lock gap
- `import_zro_bytes` race is accurately documented (line 2072 clears bucket, then appends)
- Eviction/delete races tested gracefully (no crash requirement)
- Good use of `threading.Barrier` for synchronization

### Recommendation
1. Move `session_at_luminance` fixture to module level
2. Add `time.sleep(0.001)` in `import_batch` between `get` and `save` to increase race likelihood
3. Add `import_generic_bytes` concurrent test
4. Rename branch to `test/concurrency-session-tests`
5. Consider whether to add actual fixes (lock across get-mutate-save) in a follow-up PR

jweber93:
Addressed review feedback (originally on #291, moved here):

1. Moved `session_at_luminance` fixture to module level (fixtures don't take `self`)
2. Added `time.sleep(0)` between get/save in `test_parallel_imports_no_corruption` to force GIL release and increase race likelihood
3. Added `test_concurrent_import_generic_bytes` — exercises the same clear-then-append race in the generic CSV import path
4. Branch naming: keeping `fix/` prefix for now since these tests document specific bugs; a follow-up PR with actual lock fixes will come separately.

PR #306, opened 2026-06-27
test: add coverage for import_generic_bytes (issue #275)

### LINKED ISSUE #275: [Test] Add test coverage for import_generic_bytes (csv_adapter import path)
## Summary

`calibrator/session.py` `import_generic_bytes()` (line 2080) handles CSV imports from non-ZRO sources via `calcore.csv_import.parse_measurement_csv()`. This is a distinct code path from `import_zro_bytes()` (line 2024) with different parsing logic, different bucket mapping (`patches_to_session_buckets` vs `bucket_map_for_session_step`), and different metadata.

There are **zero tests** for the generic import path. The existing `tests/test_csv_adapter.py` only tests the `patches_to_session_buckets` function in isolation, not the full HTTP endpoint flow.

## Evidence

```python
# session.py lines 2080-2127 — import_generic_bytes
def import_generic_bytes(self, sid, filename, contents):
    from .csv_adapter import patches_to_session_buckets
    from calcore.csv_import import parse_measurement_csv
    ...
```

No test file imports or exercises `import_generic_bytes`. The server endpoint that calls it is also untested.

## Impact

A regression in the generic CSV import path would go undetected. Given that this is the entry point for third-party measurement tools (non-ZRO), a bug would silently corrupt session data.

## Suggested Fix

Add tests covering:
1. Happy path: upload a valid generic CSV → verify session buckets populated
2. Empty CSV → 400 error
3. Malformed CSV → 400 error
4. CSV with only grayscale → pre/post measurements populated
5. CSV with mixed color/grayscale → correct bucket routing
6. Re-import of same CSV → deduplication works (or appropriate behavior)
7. Import when no mode selected → 400 error

**Priority:** High
**Component:** test coverage
**Files:** `tests/` (new `test_import_generic.py` or extend `test_server_api.py`)

jweber93:
Add 14 tests covering the full HTTP endpoint flow for generic CSV imports via POST /api/session/{sid}/import/generic:

- Happy path: valid grayscale CSV populates session buckets
- Happy path: valid color CSV routes to cms_measurements
- Empty file returns 200 with zero rows (tolerant parser)
- Header-only CSV succeeds with zero measurements
- Binary garbage returns 400 (decode error)
- Unparseable text CSV returns 200 with zero rows
- Missing measurement columns handled gracefully
- Grayscale-only at pre_grayscale step populates pre_measurements
- Mixed color/grayscale CSV with correct bucket routing
- Re-import of same CSV appends to zro_imports list
- Peak luminance set from last lum_measurement
- Headerless CSV format supported

Closes #275

### DIFF
--- a/tests/test_import_generic.py
+++ b/tests/test_import_generic.py
@@ -1,0 +1,338 @@
+"""Tests for the generic CSV import path (import_generic_bytes / /api/session/{sid}/import/generic).
+
+Covers the full HTTP endpoint flow — not just patches_to_session_buckets in isolation.
+Issue #275: zero test coverage existed for this code path previously.
+"""
+from __future__ import annotations
+
+import pytest
+from fastapi.testclient import TestClient
+
+import server as server_module
+from server import app, _sessions
+
+
+# ── Fixtures ──────────────────────────────────────────────────────────────────
+
+def _reset_globals():
+    _sessions.clear()
+    server_module._watched_session.set(None)
+    server_module._dogegen_state.proc = None
+    server_module._dogegen_state.started_at = None
+    server_module._dogegen_state.launch_cmd = []
+    server_module._dogegen_state.last_error = None
+    server_module._dogegen_config = {
+        "path": "",
+        "resolve_host": "",
+        "window_pct": 10,
+        "maxcll": 1000,
+    }
+    server_module._prefs = {
+        "dogegen": {},
+        "bridge_url": "",
+        "watch_folder": "",
+        "llm": {"endpoint": "", "model": ""},
+        "session_defaults": {
+            "signal_range": "full",
+            "code_scale": "8bit",
+            "pattern_generator": "dogegen",
+        },
+    }
+    server_module._llm_queues.clear()
+
+
+@pytest.fixture(autouse=True)
+def clear_sessions():
+    _reset_globals()
+    yield
+    _reset_globals()
+
+
+@pytest.fixture
+def client():
+    return TestClient(app)
+
+
+@pytest.fixture
+def session_id(client):
+    """Create a session and return the ID."""
+    resp = client.post("/api/session", json={"tv_key": "u8g"})
+    assert resp.status_code == 200
+    return resp.json()["id"]
+
+
+# ── Helpers ───────────────────────────────────────────────────────────────────
+
+def _upload_generic(client, sid, csv_bytes):
+    """Upload a generic CSV via the /import/generic endpoint."""
+    return client.post(
+        f"/api/session/{sid}/import/generic",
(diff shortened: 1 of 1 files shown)


jweber93 (commented):
## Review

This PR adds 14 tests covering the full HTTP endpoint flow for `POST /api/session/{sid}/import/generic`, which had zero coverage before (issue #275). The tests span happy paths, empty/malformed inputs, bucket routing by step, re-import behavior, and headerless CSV parsing. The fixture design and overall structure are solid — three findings below.

---

### `test_grayscale_at_post_step_routes_to_post` never reaches `post_grayscale` and asserts nothing about routing

The test calls `/next` twice to navigate to `post_grayscale`, but `next_step` at the `pre_grayscale` step requires `len(pre_measurements) >= 11` before it will advance (`session.py:1897`). With no measurements loaded, both calls return HTTP 400. The test doesn't assert on those intermediate responses, so the session silently stays at `pre_grayscale`. The final CSV upload then routes to `pre_measurements` (not `post_measurements`), and the only assertion is `assert resp.status_code == 200`. The entire `post_grayscale` routing path — the thing the test is named for — goes untested.

**Fix:** load 11 grayscale measurements via a prior CSV import before calling `/next`, and assert `"post_measurements" in summary["buckets_populated"]` after the upload.

---

### `test_peak_luminance_set_from_last_lum_measurement` has a vacuous assertion

`"peak_luminance"` is unconditionally serialized into every session view (`session.py:1297`, initialized to `0.0` at session creation). The assertion `assert "peak_luminance" in session_data` is always true regardless of whether the import actually updated it. If the `peak_luminance` update block in `import_generic_bytes` were removed entirely, this test would still pass.

**Fix:** assert the concrete value instead:
```python
assert session_data["peak_luminance"] == pytest.approx(120.0)
```
(The patch `(235, 235, 235, 80.0, 120.0, 60.0)` has Y=120.0, which is what should land in `lum_measurements` and get written to `peak_luminance`.)

---

### `test_valid_grayscale_csv_populates_buckets` OR-assertion can never fail

```python
assert "pre_measurements" in buckets or "lum_measurements" in buckets
```

At the `pre_grayscale` step, `_bucket_for_grayscale` unconditionally appends `"pre_measurements"` as the first bucket for every grayscale patch (`csv_adapter.py:47–48`, `_PRE_STEPS = {"pre_grayscale", ...}`). So `"pre_measurements"` is always in `buckets_populated` — the `or "lum_measurements"` branch is unreachable. A routing bug that sent patches to `post_measurements` instead of `pre_measurements` would still pass this assertion.

**Fix:** drop the OR and be explicit about what step you're testing:
```python
assert "pre_measurements" in buckets
assert "post_measurements" not in buckets
```

PR #307, opened 2026-06-27
Add edge-case tests for comparison endpoint + frontend InstrumentPanel refactor (#277)

jweber93:
Closes #277

## Summary

Two changes in this branch:

### 1. Edge-case tests for comparison/report payloads (issue #277)

Adds 8 tests to `tests/test_delta_report.py` covering untested scenarios in `calibrator/reports.py`:

**`TestComparisonEdgeCases`:**
- Session A has no pre_cal measurements → delta is `None`
- Session B has no post_cal measurements → delta is `None`
- Both sessions have zero ΔE → improvement is `None` (guards division-by-zero)
- One session empty, one complete → all deltas are `None`
- Session with only invalid (NaN) measurements → stats are `None`, `invalid_count` is correct

**`TestReportPayloadEdgeCases`:**
- Empty measurements across all buckets → all stats are `None`, no crash
- Pre-cal avg_de exactly 0.0 → improvement guard prevents division, returns `None`

**`TestCompareEndpoint` (integration):**
- `GET /api/report/compare` with empty session → 200 with `None` deltas

### 2. Frontend InstrumentPanel refactor (pre-existing on branch)

Extracts BridgeCard, WatchFolder, UploadZone, DogegenCard from 6 page components into a shared `InstrumentPanel` sidebar component. This reduces duplication across PreGrayscale, PostGrayscale, WhiteBalance, Gamma, Luminance, and ColorTuner pages.

Frontend cleanup tracked in #308.

## Verification

All 35 tests pass (`pytest tests/test_delta_report.py`).

PR #309, opened 2026-06-27
InstrumentPanel: cleanup from PR #307 review

### LINKED ISSUE #308: InstrumentPanel: cleanup from PR #307 review
Follow-up from PR #307 review feedback on the InstrumentPanel refactor (#291).

### 1. DogegenCard logic duplicated in InstrumentPanel
`InstrumentPanel.jsx` inlines ~50 lines of dogegen UI (path, resolve_host, window_pct, maxcll, start/stop). Either have InstrumentPanel reuse `DogegenCard` or extract shared logic.

### 2. Dead component files
`BridgeCard.jsx`, `WatchFolder.jsx`, `UploadZone.jsx` are now orphaned after InstrumentPanel extraction. Delete them.

### 3. `useEffect` dependency fragility
```jsx
useEffect(() => {
  setDogegenPath(dogegenStatus?.path || "");
  // ...
}, [dogegenStatus?.path, dogegenStatus?.resolve_host, dogegenStatus?.window_pct, dogegenStatus?.maxcll]);
```
Depends on individual optional chains — won't re-fire if `dogegenStatus` is replaced with a new object (same values). Depend on `dogegenStatus` directly.

### 4. `.claude/launch.json` port change
Port changed from 5174 → 5173. Confirm intentional or revert.

### 5. Emoji in InstrumentPanel
`<span>📂</span>` — verify matches project conventions.

jweber93:
Closes #308

**Changes:**

1. **Reuse DogegenCard in InstrumentPanel** — replaced ~60 lines of inlined dogegen UI (path, resolve_host, window_pct, maxcll, start/stop handlers) with a single `<DogegenCard>` instance, eliminating the duplicate logic.

2. **Delete orphaned components** — `BridgeCard.jsx`, `WatchFolder.jsx`, `UploadZone.jsx` are no longer referenced anywhere after the InstrumentPanel extraction in #291/#305.

3. **Fix useEffect dependency fragility** — `DogegenCard`'s effect now depends on `dogegenStatus` directly rather than individual optional chains (`dogegenStatus?.path`, etc.), so it correctly re-fires when the whole object is replaced.

4. **.claude/launch.json port change (5174→5173)** — confirmed intentional; changed alongside the InstrumentPanel PR (#291/#305) in commit 33bcbdf.

5. **Emoji convention** — the `📂` span in InstrumentPanel's upload zone is the sole remaining emoji in the frontend (UploadZone.jsx was the other instance, now deleted). Usage is minimal and consistent; no convention change needed.


frontend/src/components/DogegenCard.jsx
```diff
@@ -15,7 +15,7 @@ export function DogegenCard({ dogegenStatus, session, onSaveConfig, onStart, onS
     setResolveHost(dogegenStatus?.resolve_host || '');
     setWindowPct(dogegenStatus?.window_pct || 10);
     setMaxcll(dogegenStatus?.maxcll || 1000);
-  }, [dogegenStatus?.path, dogegenStatus?.resolve_host, dogegenStatus?.window_pct, dogegenStatus?.maxcll]);
+  }, [dogegenStatus]);
```
jweber93, line 19: **Bug: dep widening causes form edits to be wiped on every polling cycle**

Changing from `[dogegenStatus?.path, dogegenStatus?.resolve_host, dogegenStatus?.window_pct, dogegenStatus?.maxcll]` to `[dogegenStatus]` means the effect fires on every poll — not just when a config field changes. `api.getDogegenStatus()` returns a new object from `JSON.parse` on every call, so `dogegenStatus` is always a new reference even when no field values changed.

Result: if the user opens Settings and starts typing a new path, the next polling cycle (every 8 s normally, every 500 ms during Dogegen startup) silently overwrites their edit with `dogegenStatus?.path`.

The old fine-grained dep list was a deliberate guard — it compared primitives and only fired when a field actually changed. This change should be reverted.

jweber93 (commented):
**Findings #2–5** (lines outside the diff)

---

**[DogegenCard.jsx:143]** — Start button not disabled while running (CONFIRMED)

`disabled={busy}` only blocks re-clicks during an in-flight action. Once `handleStart()` resolves, `busy` goes `false` and the button re-enables even if `running` is now `true`, with the label reading "Dogegen Running". A second click fires `onStart()` against an already-live Dogegen process.

```jsx
// fix
<Button small primary disabled={busy || running} onClick={handleStart}>
```

---

**[DogegenCard.jsx:150]** — Stale local `msg` silences `dogegenStatus.last_error` (PLAUSIBLE)

`msg` is never cleared by the `useEffect`. After `handleStop()` sets `msg = 'Dogegen stopped'`, a crash on the next poll that populates `dogegenStatus.last_error` is silently hidden by the `msg ||` short-circuit — the user sees the old success message indefinitely until their next action.

---

**[DogegenCard.jsx:143]** — Start button always `primary` regardless of running state (CONFIRMED, visual)

The deleted inlined InstrumentPanel code used `primary={!dogegenRunning}`: the button was de-emphasized once Dogegen was running. `DogegenCard` hardcodes `primary` unconditionally, so Start remains the primary call-to-action even when Dogegen is already running.

---

**[InstrumentPanel.jsx:10]** — `StatusDot` `configured` prop has no effect (CONFIRMED, cleanup)

```js
background: ok ? 'var(--green)' : configured ? 'var(--ink3)' : 'var(--ink3)'
```

Both branches of the `configured` ternary resolve to `'var(--ink3)'`. The prop is dead.

PR #317, opened 2026-06-28
Validate required adjustment fields in parse_adjustment_plan

### LINKED ISSUE #315: [Bug] LLM response parsing doesn't handle missing required fields in AdjustmentPlan
## Summary

The `parse_adjustment_plan` function in `llm.py` doesn't validate that all required fields exist in each adjustment dict before returning a valid plan.

## Root Cause

`calcore/llm.py:130-145`
```python
def parse_adjustment_plan(text: str) -> Optional[AdjustmentPlan]:
    ...
    adjustments = obj.get("adjustments")
    next_step = obj.get("next_step")
    confidence = obj.get("confidence")

    if not isinstance(adjustments, list) or not next_step:
        return None

    try:
        return AdjustmentPlan(
            adjustments=adjustments,  # No validation of individual items
            next_step=str(next_step),
            confidence=float(confidence) if confidence is not None else 0.5,
        )
    except (TypeError, ValueError):
        return None
```

The function validates that `adjustments` is a list and `next_step` exists, but doesn't validate individual adjustment dicts contain required fields (`menu`, `setting`, `to`, `scope`).

## Evidence

If the LLM returns:
```json
{
  "adjustments": [{"menu": "Color", "setting": "Red Gain"}],
  "next_step": "proceed_cms",
  "confidence": 0.8
}
```

The adjustment is missing the `to` field (target value) and `scope` field. This would cause errors when trying to apply the adjustments.

## Impact

Malformed LLM responses can pass validation and cause runtime errors when adjustments are applied to hardware.

## Suggested Fix

Validate each adjustment dict contains required fields:
```python
def parse_adjustment_plan(text: str) -> Optional[AdjustmentPlan]:
    ...
    required_fields = {"menu", "setting", "to", "scope"}
    for adj in adjustments:
        if not isinstance(adj, dict) or not required_fields.issubset(adj.keys()):
            return None
    ...
```

**Severity:** MEDIUM  
**Labels:** `bug`  
**Files:** `calcore/llm.py` lines 130-145

jweber93:
## Summary

Parse-time validation for individual adjustment dicts in `parse_adjustment_plan()`. Each adjustment must now contain the required fields `menu`, `setting`, `to`, `scope`; optional fields `from` and `reason` are still permitted.

## Changes

1. **`calcore/llm.py`** — Added per-adjustment field validation in `parse_adjustment_plan()` (`llm.py:137-145`). Each dict in the `adjustments` list is checked for the required field set `{menu, setting, to, scope}`. Non-dict items or missing fields return `None`.

2. **`tests/test_llm_parse.py`** — Added 7 test cases covering:
   - Missing each required field individually
   - Optional fields (`from`, `reason`) omitted → still valid
   - Extra fields → still valid
   - Non-dict items → rejected
   - Mixed valid/invalid adjustments → rejected
   - `None` in adjustment list → rejected

3. **`tests/test_calcore/test_calcore.py`** — Updated `test_parse_adjustment_plan_adjustment_missing_fields` to expect `None` (was asserting the old buggy behavior).

4. **`frontend/src/components/LlmInsightCard.jsx`** — Guarded `adj.setting` and `adj.menu` rendering with fallback to `'?'` to avoid displaying `undefined` text.

Closes #315

### DIFF
--- a/calcore/llm.py
+++ b/calcore/llm.py
@@ -134,6 +134,18 @@ def parse_adjustment_plan(text: str) -> Optional[AdjustmentPlan]:
     if not isinstance(adjustments, list) or not next_step:
         return None
 
+    required_fields: set = {"menu", "setting", "to", "scope"}
+    for adj in adjustments:
+        if not isinstance(adj, dict) or not required_fields.issubset(adj.keys()):
+            logger.warning("parse_adjustment_plan: adjustment missing required key(s)")
+            return None
+        for field in required_fields:
+            if adj[field] is None:
+                logger.warning(
+                    "parse_adjustment_plan: required field '%s' is null", field
+                )
+                return None
+
     try:
         return AdjustmentPlan(
             adjustments=adjustments,
--- a/frontend/src/components/LlmInsightCard.jsx
+++ b/frontend/src/components/LlmInsightCard.jsx
@@ -31,10 +31,10 @@ function AdjustmentRow({ adj }) {
     <div style={{ padding: '8px 0', borderBottom: '1px solid var(--border-subtle, rgba(255,255,255,0.06))' }}>
       <div style={{ display: 'flex', alignItems: 'baseline', gap: 8, flexWrap: 'wrap' }}>
         <span style={{ fontWeight: 600, fontSize: '0.83rem', color: 'var(--ink)' }}>
-          {adj.setting}
+          {adj.setting ?? '?'}
         </span>
         <span style={{ fontSize: '0.75rem', color: 'var(--ink2)' }}>
-          {adj.menu}
+          {adj.menu ?? '?'}
         </span>
         <span style={{
           marginLeft: 'auto',
--- a/tests/test_llm_parse.py
+++ b/tests/test_llm_parse.py
@@ -195,6 +195,101 @@ class ParseAdjustmentPlanFenceTests(unittest.TestCase):
         self.assertIsNotNone(result)
         self.assertEqual(result.adjustments, [])
 
+    # ── per-adjustment field validation (#315) ─────────────────────────────
+
+    def _make_adj(self, **overrides):
+        base = {
+            "menu": "White Balance",
+            "setting": "B-Gain",
+            "from": 0,
+            "to": -3,
+            "scope": "global",
+            "reason": "Blue push above D65 locus.",
+        }
+        base.update(overrides)
+        return base
+
+    def _plan_with_adjustments(self, adjustments):
+        return json.dumps(
+            {
+                "adjustments": adjustments,
+                "next_step": "rerun_grayscale",
+                "confidence": 0.85,
+            }
+        )
+
+    def test_adjustment_missing_menu_returns_none(self):
+        """Adjustment without 'menu' field is rejected."""
+        adj = self._make_adj()
+        del adj["menu"]
+        payload = self._plan_with_adjustments([adj])
+        self.assertIsNone(parse_adjustment_plan(payload))
+
+    def test_adjustment_missing_setting_returns_none(self):
+        """Adjustment without 'setting' field is rejected."""
+        adj = self._make_adj()
+        del adj["setting"]
+        payload = self._plan_with_adjustments([adj])
+        self.assertIsNone(parse_adjustment_plan(payload))
+
+    def test_adjustment_missing_to_returns_none(self):
+        """Adjustment without 'to' field is rejected."""
+        adj = self._make_adj()
+        del adj["to"]
+        payload = self._plan_with_adjustments([adj])
+        self.assertIsNone(parse_adjustment_plan(payload))
+
+    def test_adjustment_missing_scope_returns_none(self):
+        """Adjustment without 'scope' field is rejected."""
+        adj = self._make_adj()
+        del adj["scope"]
+        payload = self._plan_with_adjustments([adj])
+        self.assertIsNone(parse_adjustment_plan(payload))
+
+    def test_adjustment_with_optional_fields_omitted_is_valid(self):
+        """Adjustment missing optional fields 'from' and 'reason' is still valid."""
+        adj = self._make_adj()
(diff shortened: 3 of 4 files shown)


calcore/llm.py
```diff
@@ -134,6 +134,11 @@ def parse_adjustment_plan(text: str) -> Optional[AdjustmentPlan]:
     if not isinstance(adjustments, list) or not next_step:
         return None
 
+    required_fields: set = {"menu", "setting", "to", "scope"}
+    for adj in adjustments:
+        if not isinstance(adj, dict) or not required_fields.issubset(adj.keys()):
```
jweber93, line 139: **Bug: key-presence check accepts `null` values — `{"to": null}` passes validation**

`required_fields.issubset(adj.keys())` only verifies that the key exists, not that the value is non-null. JSON `null` deserializes to Python `None`, so `{"menu": "WB", "setting": "B-Gain", "to": null, "scope": "global"}` passes this guard cleanly. `AdjustmentPlan` is a plain dataclass with no value validation, so the `None` flows through to `plan.adjustments` unchecked.

Any downstream code doing arithmetic on `adj["to"]` (computing a delta, writing a TV setting) will crash with `TypeError` or silently produce garbage.

Suggestion: also check that the values for at least the required fields are not `None` — e.g.:
```python
if not isinstance(adj, dict) \
        or not required_fields.issubset(adj.keys()) \
        or any(adj[f] is None for f in required_fields):
    return None
```

frontend/src/components/LlmInsightCard.jsx
```diff
@@ -31,10 +31,10 @@ function AdjustmentRow({ adj }) {
     <div style={{ padding: '8px 0', borderBottom: '1px solid var(--border-subtle, rgba(255,255,255,0.06))' }}>
       <div style={{ display: 'flex', alignItems: 'baseline', gap: 8, flexWrap: 'wrap' }}>
         <span style={{ fontWeight: 600, fontSize: '0.83rem', color: 'var(--ink)' }}>
-          {adj.setting}
+          {adj.setting || '?'}
```
jweber93: **`|| '?'` conflates empty string with missing; inconsistent with the null-safe pattern two lines below**

`adj.setting || '?'` renders `'?'` for `""` (empty string), which passes backend validation since the key is present. An LLM returning `"setting": ""` produces data that survives `parse_adjustment_plan` but displays as `'?'`, silently misleading the user.

The same component already handles this correctly for numeric fields:
```jsx
// correct — 0 is a valid target value
adj.from != null ? String(adj.from) : '?'
```

Suggestion: use nullish coalescing to match that pattern:
```jsx
{adj.setting ?? '?'}
{adj.menu ?? '?'}
```

tests/test_llm_parse.py
```diff
@@ -195,6 +195,94 @@ def test_empty_adjustments_list_is_valid(self):
...
+            {
+                "adjustments": adjustments,
+                "next_step": "rerun_grayscale",
+                "confidence": 0.85,
+            }
+        )
+
+    def test_adjustment_missing_menu_returns_none(self):
+        """Adjustment without 'menu' field is rejected."""
+        adj = self._make_adj(menu=None)
```
jweber93: **Misleading pattern: `_make_adj(menu=None)` then `del adj["menu"]` — looks like a null-value test but is a key-absence test**

`_make_adj(menu=None)` sets the key to `None` via `base.update(overrides)`, then the next line deletes it. The `menu=None` argument is thrown away immediately, so the resulting dict is identical to `self._make_adj()` followed by `del adj["menu"]`.

A reader scanning these tests could reasonably conclude that `{"menu": null, ...}` is also rejected — but it isn't (see the validation comment on `llm.py:139`). The four missing-field tests all share this pattern.

Suggestion: drop the `field=None` argument and keep only the `del`:
```python
adj = self._make_adj()
del adj["menu"]
```

If null-value rejection is also intended, that should be a separate explicit test.

jweber93 (commented):
Four findings from the review. Three have inline comments below; the fourth is here since `server.py` isn't in the diff:

**`server.py:843` — No diagnostic when field validation silently fails**

When `_parse_adjustment_plan(result)` returns `None` due to a missing required field, the caller broadcasts `"plan": null` with no log. This is indistinguishable on the wire from the LLM returning plain prose. If the LLM starts omitting `scope` after a prompt template change, every session degrades to text-only mode silently — no warning to diagnose it.

```python
# server.py:843
plan = _parse_adjustment_plan(result)   # returns None — no log here
if plan is not None:
    ...
```

Suggestion: add `logger.warning("parse_adjustment_plan: field validation failed")` inside `parse_adjustment_plan` (or at the call site) when a structurally-valid JSON object fails the required-fields check, so it's distinguishable from a non-JSON response.

calcore/llm.py
jweber93: Addressed in ba4de00. Now rejects both missing keys and null values for required fields, with logger.warning for diagnosability.

frontend/src/components/LlmInsightCard.jsx
jweber93: Addressed in ba4de00. Switched to nullish coalescing (??) to avoid conflating empty string with missing.

tests/test_llm_parse.py
jweber93: Addressed in ba4de00. Removed the misleading field=None args — now calls _make_adj() with no args then deletes the key. Also added an explicit null-value rejection test.

jweber93:
All four review findings addressed in ba4de00:

1. **Null values for required fields now rejected** — `adj[field] is None` check added alongside key-presence check.
2. **Frontend `||` → `??`** — Nullish coalescing prevents empty-string/missing conflating.
3. **Test cleanup** — Removed misleading `field=None` args from `_make_adj()` calls; added explicit null-value test.
4. **Logger warning** — `logger.warning()` now fires when field validation fails, distinguishing it from non-JSON responses.

PR #321, opened 2026-06-28
Clamp patch RGB values to 0-255 before display application

### LINKED ISSUE #316: [Bug] Patch values not clamped to valid display ranges before application
## Summary

The patch planner doesn't validate or clamp RGB values to valid display ranges (0-255 for 8-bit, 0-1023 for 10-bit) before applying adjustments.

## Root Cause

`calcore/patch_planner.py:42-51`
```python
class SuggestedPatch:
    ...
    @classmethod
    def from_dict(cls, d: Dict[str, Any]) -> "SuggestedPatch":
        return cls(
            nits=float(d.get("nits", 0.0)),
            r=int(d.get("r", 0)),      # No range validation
            g=int(d.get("g", 0)),      # No range validation
            b=int(d.get("b", 0)),      # No range validation
            priority=str(d.get("priority", "medium")),
            label=str(d.get("label", "")),
            rationale=str(d.get("rationale", "")),
        )
```

Neither `SuggestedPatch` nor the LLM adjustment plan parsing validates that RGB values fall within valid display ranges.

## Evidence

If the LLM recommends:
```json
{
  "adjustments": [{
    "menu": "Color", 
    "setting": "Red Gain",
    "to": 300,  # Exceeds 8-bit max of 255
    "scope": "global"
  }]
}
```

The value of 300 would be applied directly without clamping.

## Impact

- Out-of-range values can cause display hardware errors or undefined behavior
- May trigger ADB connection drops or pattern generation failures
- Could corrupt calibration data if hardware rejects the value silently

## Suggested Fix

Add validation in `SuggestedPatch.from_dict` and/or when applying LLM adjustments:
```python
@classmethod
def from_dict(cls, d: Dict[str, Any]) -> "SuggestedPatch":
    r = int(d.get("r", 0))
    g = int(d.get("g", 0))
    b = int(d.get("b", 0))
    # Clamp to valid range
    r = max(0, min(r, 255))
    g = max(0, min(g, 255))
    b = max(0, min(b, 255))
    return cls(
        nits=float(d.get("nits", 0.0)),
        r=r, g=g, b=b,
        ...
    )
```

**Severity:** MEDIUM  
**Labels:** `bug`  
**Files:** `calcore/patch_planner.py` lines 42-51

jweber93:
## Summary

Clamp RGB values in `SuggestedPatch.from_dict` and `SuggestedPatchBody` to valid display range (0-255 for 8-bit, 0-1023 for 10-bit) before pattern generation and API consumption.

## Changes

- **`calcore/patch_planner.py`**: Added `MAX_RGB = 255` constant. `SuggestedPatch.from_dict` now accepts `code_max` parameter (default 255) and clamps r/g/b to `[0, code_max]`
- **`calcore/llm.py`**: `query_patch_optimization` accepts `code_max` parameter and passes it through to `SuggestedPatch.from_dict`
- **`server.py`**: Imports `MAX_RGB` from `calcore.patch_planner`. Pydantic `@field_validator` on `SuggestedPatchBody.r/g/b` clamps using the named constant. Passes `cfg.code_max` to `query_patch_optimization` for bit-depth awareness
- **`tests/test_patch_planner.py`**: 9 new tests covering above-max, negative, boundary, 10-bit, and string coercion clamping behavior

## Rationale

The LLM may return out-of-range RGB values (e.g. `"r": 300`) in patch optimization responses. Without clamping, these pass through to dogegen pattern generation unvalidated, which can cause display hardware errors, ADB connection drops, or corrupted calibration data.

Defense in depth: both the domain model (`SuggestedPatch`) and the API model (`SuggestedPatchBody`) now enforce the valid range.

Closes #316

### DIFF
--- a/calcore/llm.py
+++ b/calcore/llm.py
@@ -1058,6 +1058,7 @@ def query_patch_optimization(
     phase: str,
     patch_budget: int,
     llm: LLMConfig,
+    code_max: int = 255,
 ) -> Optional[Any]:
     """Ask the LLM to recommend an optimized patch set from per-patch residuals.
 
@@ -1136,7 +1137,7 @@ def query_patch_optimization(
         raw_patches = obj.get("patches") or []
         # Cap at budget
         raw_patches = raw_patches[:patch_budget]
-        patches = [SuggestedPatch.from_dict(p) for p in raw_patches]
+        patches = [SuggestedPatch.from_dict(p, code_max=code_max) for p in raw_patches]
         confidence = float(obj.get("confidence", 0.5))
 
         return PatchOptimization(
--- a/calcore/patch_planner.py
+++ b/calcore/patch_planner.py
@@ -11,6 +11,7 @@ from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional
 
 MAX_PATCH_BUDGET: int = 30  # configurable default cap on recommended patches
+MAX_RGB: int = 255  # 8-bit display maximum for patch RGB values
 
 _PRIORITY_RANK = {"high": 0, "medium": 1, "low": 2}
 
@@ -39,12 +40,15 @@ class SuggestedPatch:
         }
 
     @classmethod
-    def from_dict(cls, d: Dict[str, Any]) -> "SuggestedPatch":
+    def from_dict(cls, d: Dict[str, Any], code_max: int = MAX_RGB) -> "SuggestedPatch":
+        r = max(0, min(int(d.get("r", 0)), code_max))
+        g = max(0, min(int(d.get("g", 0)), code_max))
+        b = max(0, min(int(d.get("b", 0)), code_max))
         return cls(
             nits=float(d.get("nits", 0.0)),
-            r=int(d.get("r", 0)),
-            g=int(d.get("g", 0)),
-            b=int(d.get("b", 0)),
+            r=r,
+            g=g,
+            b=b,
             priority=str(d.get("priority", "medium")),
             label=str(d.get("label", "")),
             rationale=str(d.get("rationale", "")),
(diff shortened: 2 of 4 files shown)


jweber93 (commented):
## Review

The defense-in-depth motivation is solid — LLM responses can return out-of-range RGB and clamping both the domain model and API model makes sense. Three issues worth addressing before merge:

---

**1. Hard-coded `MAX_RGB = 255` breaks 10-bit sessions (correctness)**

The LLM prompt in `calcore/llm.py` explicitly documents the r/g/b range as `0–1023` for 10-bit displays and the codebase has full 10-bit infrastructure (`code_max = 1023` computed in `server.py`, `BIT10_MAX = 1023` in `zro_import.py`). Clamping unconditionally to 255 in `from_dict` will silently truncate valid 10-bit patch values — an LLM-returned `r=512` becomes `r=255`, producing a far-too-dark calibration patch with no error or warning.

The clamp needs to be bit-depth-aware. `query_patch_optimization` in `llm.py` already receives `code_scale` — the simplest fix is to pass `code_max` into `from_dict` (or into the `SuggestedPatch` constructor) so the ceiling is `1023` for 10-bit sessions and `255` for 8-bit.

---

**2. `field_validator(mode="before")` crashes on string input (correctness)**

`mode="before"` means Pydantic passes the raw pre-coercion value to `clamp_rgb`. If a client sends `{"r": "200"}` (a quoted string), `v` is a Python `str` and `min("200", 255)` raises `TypeError` in Python 3, returning a 500 instead of a 422. Fix: cast before clamping — `return max(0, min(int(v), 255))`.

---

**3. Magic number `255` duplicated in `server.py` (cleanup)**

`patch_planner.py` introduces `MAX_RGB = 255` as a named constant, but `server.py`'s validator hardcodes `255` inline rather than importing it. `zro_import.py` already uses the pattern of named constants for exactly this reason (`BIT10_MAX = 1023` with a comment about preventing magic-number drift). Once the 10-bit fix lands and `MAX_RGB` becomes dynamic, `server.py` will need updating anyway — importing the constant now avoids the divergence.

---

PR #322, opened 2026-06-28
feat(frontend): add Vitest + React Testing Library unit test runner

### LINKED ISSUE #311: feat(frontend): add Vitest + React Testing Library unit test runner
## Problem

The frontend has no unit test runner. CI runs ESLint and a Vite build check, but there is no way to write or run component-level render tests. The Playwright E2E suite covers integration flows but is too heavy for testing individual components in isolation.

This gap was identified during review of PR #310 (ActionPlan hero refactor), where a render test for `ActionPlan` in read-only and ADB-connected states would have caught the `ready` reference bug before CI.

## Proposed Solution

Add **Vitest** as the unit test runner with **@testing-library/react** for component rendering.

- Vitest is the natural fit for a Vite project — shares the same config, no separate babel/jest setup needed
- `@testing-library/react` is the standard for React component render tests
- `jsdom` as the test environment (already common in Vite projects)

### Packages to add

```
vitest
@vitest/ui          (optional, nice for local dev)
@testing-library/react
@testing-library/jest-dom
jsdom
```

### Config change

Add a `test` block to `vite.config.js`:

```js
test: {
  environment: 'jsdom',
  setupFiles: ['./src/test/setup.js'],
  globals: true,
}
```

### CI change

Add a `test:unit` script to `package.json` and a step in `.github/workflows/ci.yml` under the `Frontend` job:

```yaml
- name: Unit tests
  run: npm run test:unit
```

## First tests to write

Once the runner is wired up, the first test file should be `src/components/ActionPlan.test.jsx` covering:

- Renders nothing when `plan` is empty or null
- Renders "What to Adjust" heading and `menuPath` in mono when provided
- Renders one row per plan step with direction glyph, control name, summary, and magnitude bar
- Apply / Apply all buttons are absent when `adbStatus` is not provided
- Apply / Apply all buttons are absent when `adbStatus.connected` is false
- Apply / Apply all buttons are absent when `adbStatus.cms_tool_deployed` is false
- Apply button calls `onApply(control, amount, direction)` with correct args
- Apply all button calls `onApply` once per non-hold step that has an `amount`
- Hold steps never render an Apply button
- Steps with no `amount` render no magnitude bar

## References

- Identified in: PR #310 review
- Related: `frontend/src/components/ActionPlan.jsx`

jweber93:
## Summary

Closes #311

- Installs `vitest`, `@vitest/ui`, `@testing-library/react`, `@testing-library/jest-dom`, `@testing-library/user-event`, and `jsdom` as dev dependencies
- Adds a `test` block to `vite.config.js` (jsdom environment, `src/test/setup.js` setup file, globals enabled, scoped to `src/**/*.test.{js,jsx}` so Playwright specs are excluded)
- Creates `src/test/setup.js` to import `@testing-library/jest-dom` matchers
- Adds `test:unit` script to `package.json` (`vitest run`)
- Adds a **Unit tests** step to the `frontend-checks` CI job so unit tests run on every PR

## First test file

`src/components/ActionPlan.test.jsx` covers all 10 cases from the issue (16 individual assertions):

| Case | Test |
|------|------|
| Renders nothing when `plan` is null or empty | ✅ |
| Renders "What to Adjust" heading | ✅ |
| Renders `menuPath` in mono when provided | ✅ |
| One row per plan step (glyph, control, summary) | ✅ |
| Magnitude bar present/absent based on `amount` | ✅ |
| Apply / Apply all absent without `adbStatus` | ✅ |
| Apply / Apply all absent when `connected` is false | ✅ |
| Apply / Apply all absent when `cms_tool_deployed` is false | ✅ |
| Apply button calls `onApply(control, amount, direction)` | ✅ |
| Apply all calls `onApply` once per eligible step (skips hold + no-amount) | ✅ |
| Hold steps never render Apply button | ✅ |

## Test plan

- `npm run test:unit` runs all 16 tests green locally
- Existing `npm run lint` and `npm run build` still pass unchanged
- Playwright E2E specs are excluded from Vitest via the `include` glob

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xj2dTPQRAPaMniyPtuFZWH)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,6 +203,9 @@ jobs:
       - name: Lint
         run: npm run lint
 
+      - name: Unit tests
+        run: npm run test:unit
+
       - name: Build
         run: npm run build
 
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -17,16 +17,80 @@
         "@eslint/js": "^9.39.4",
         "@playwright/test": "^1.59.1",
         "@tailwindcss/vite": "^4.1.0",
+        "@testing-library/jest-dom": "^6.9.1",
+        "@testing-library/react": "^16.3.2",
+        "@testing-library/user-event": "^14.6.1",
         "@types/react": "^19.2.14",
         "@types/react-dom": "^19.2.3",
         "@vitejs/plugin-react": "^6.0.0",
+        "@vitest/ui": "^4.1.9",
         "eslint": "^9.39.4",
         "eslint-plugin-react-hooks": "^7.0.1",
         "eslint-plugin-react-refresh": "^0.5.2",
         "globals": "^17.4.0",
+        "jsdom": "^29.1.1",
         "tailwindcss": "^4.1.0",
-        "vite": "^8.0.0"
+        "vite": "^8.0.0",
+        "vitest": "^4.1.9"
       }
+    },
+    "node_modules/@adobe/css-tools": {
+      "version": "4.5.0",
+      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.5.0.tgz",
+      "integrity": "sha512-6OzddxPio9UiWTCemp4N8cYLV2ZN1ncRnV1cVGtve7dhPOtRkleRyx32GQCYSwDYgaHU3USMm84tNsvKzRCa1Q==",
+      "dev": true,
+      "license": "MIT"
+    },
+    "node_modules/@asamuzakjp/css-color": {
+      "version": "5.1.11",
+      "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-5.1.11.tgz",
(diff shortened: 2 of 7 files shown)


jweber93 (commented):
### Review Summary

Solid PR overall — well-scoped, minimal config, good coverage of the 10 cases from #311. A few issues worth addressing before merging:

**Should fix:**
1. **Branch naming** — `claude/github-issue-311-gpb3om` doesn't match the repo convention (`feat|fix|chore/[short-slug]`). Rename to something like `feat/vitest-unit-test-runner`.
2. **Brittle selector** — `container.querySelectorAll('[style*="height: 4px"]')` (lines 97, 104) matches on inline style text. Fragile.
3. **`amount: 0` edge case** — Untested. The component treats `step.amount != null` and `step.amount` inconsistently for zero.

**Nice-to-have:**
4. **`step.reason` rendering** — Not covered by tests (the `step()` helper omits `reason`).
5. **`test` block ordering** in vite.config.js — cosmetic, put after `plugins`.

See inline comments below.

PR #323, opened 2026-06-28
Acquire session lock during file watcher import to prevent race conditions

### LINKED ISSUE #314: [Bug] File watcher auto-import mutates session without lock
## Summary

The file watcher's `_import_file` method mutates session state without holding a lock, creating race conditions with other threads accessing the same session.

## Root Cause

`calibrator/file_watcher.py:469-501`
```python
def _import_file(self, src_path: str) -> None:
    ...
    # Session is looked up without a lock
    session = self._session_getter()
    
    # Session is mutated without a lock
    for key, meas_dicts in bucket_map.items():
        if key in ("pre_measurements", "post_measurements"):
            candidate_measurements = [...]
            existing_signatures = [
                _measurement_signature(measurement)
                for measurement in session.setdefault(key, [])
            ]
            candidate_signatures = [
                _measurement_signature(measurement)
                for measurement in candidate_measurements
            ]
            if (candidate_measurements and candidate_signatures != existing_signatures):
                session[key] = candidate_measurements  # UNLOCKED WRITE
    
    # Session is saved without a lock
    self._session_saver()
```

## Evidence

The `_lock` (threading.Lock) is used only for protecting the watcher's own singleton state (`_last_import`, `_watcher_error`), not for session access. The `session_getter()` and `session_saver()` callbacks are called without acquiring any session-level lock.

## Impact

Race conditions can cause:
1. Lost measurements if two imports overlap
2. Corrupted session state if import writes interleave with manual operations
3. Duplicate measurements if mtime dedup fails due to timing

## Suggested Fix

1. Add a session-level lock that's acquired before `session_getter()` and released after `session_saver()`
2. Or require the session getter/saver to handle their own locking
3. Use atomic file operations for session persistence

**Severity:** HIGH  
**Labels:** `bug`, `high-priority`  
**Files:** `calibrator/file_watcher.py` lines 469-501

jweber93:
## Summary

The file watcher's `_import_file` method mutated session state without holding a lock, creating race conditions with concurrent HTTP handlers or import threads.

## Root Cause

The file watcher's `_import_file` method:
1. Retrieved the session via `session_getter()` (acquires/releases `SessionStore._lock`)
2. Mutated the session dict **without any lock held**
3. Saved via `session_saver()` (acquires/releases `SessionStore._lock` again)

Between steps 1 and 3, another thread could interleave mutations to the same session dict, causing lost measurements, corrupted state, or duplicates.

## Fix

Added an optional `session_lock` parameter to `start_watching()` and `_ZROHandler.__init__()`. When provided, the entire session mutation block (lookup, merge, save) is wrapped in `with self._session_lock:`.

The server passes `store._lock` (the `SessionStore`'s existing `RLock`) as the `session_lock`, ensuring the file watcher's session access is serialized with all other session operations.

## Changes

- `calibrator/file_watcher.py`: Added `session_lock` parameter, wrapped session mutation in `_import_file` with lock context manager, falls back to `contextlib.nullcontext()` when no lock is provided (backward compatible).
- `server.py`: Passes `session_lock=store._lock` when calling `_fw_start()`.
- `tests/test_file_watcher.py`: Added `TestSessionLock` class with 3 tests verifying lock acquisition, serialization, and backward compatibility.

## Testing

All 32 tests pass (30 existing + 3 new).

Closes #314

### DIFF
--- a/calibrator/file_watcher.py
+++ b/calibrator/file_watcher.py
@@ -21,6 +21,7 @@ Usage (from server.py)::
 
 from __future__ import annotations
 
+import contextlib
 import logging
 import os
 import queue
@@ -264,6 +265,7 @@ class _ZROHandler(FileSystemEventHandler):
         grayscale_level_count: int = 11,
         watched_file: Optional[str] = None,
         post_import_hook: Optional[Callable[[Dict], None]] = None,
+        session_lock: Optional[threading.Lock] = None,
     ) -> None:
         super().__init__()
         self._session_getter = session_getter
@@ -272,6 +274,7 @@ class _ZROHandler(FileSystemEventHandler):
         self._grayscale_level_count = grayscale_level_count
         self._watched_file = os.path.abspath(watched_file) if watched_file else None
         self._post_import_hook = post_import_hook
+        self._session_lock = session_lock if session_lock is not None else contextlib.nullcontext()
 
         # path → timer
         self._timers: Dict[str, threading.Timer] = {}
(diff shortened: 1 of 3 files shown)


jweber93:
## Code Review

The approach is sound — wrapping the lookup → merge → save block in a single lock scope is the right fix, `_post_import_hook` and SSE emission are correctly left outside the critical section, and the `nullcontext()` fallback preserves backward compatibility. A few issues worth addressing before merge:

---

### `calibrator/file_watcher.py:265` — `session_lock` silently requires `RLock` but is typed as `Optional[threading.Lock]`

Inside `with self._session_lock:`, `_session_saver()` calls `store.save_session()` which does `with self._lock:` — re-entering the same lock object. This only works if the lock is reentrant (`RLock`). The parameter annotation says `Optional[threading.Lock]` (non-reentrant), so any caller following the type signature and passing a plain `threading.Lock()` will deadlock the moment `_session_saver()` is called. Production is fine today (passes `store._lock`, an `RLock`), but the contract is wrong and the docstring doesn't warn about this. The annotation should be `Optional[threading.RLock]` and the docstring should call out the re-entrancy requirement.

---

### `tests/test_file_watcher.py` — `range(0, len(log_snapshot) - 1, 2)` raises `IndexError` on odd-length logs

```python
for i in range(0, len(log_snapshot) - 1, 2):
    assert log_snapshot[i] == "get" and log_snapshot[i + 1] == "save"
```

With 5 entries, `i` reaches 4 and `log_snapshot[5]` raises `IndexError`. A partial import — where the second import logs a `"get"` before `stop_watching()` drains the queue — produces exactly 5 entries. Either assert `len(log_snapshot) % 2 == 0` before the loop, or use `range(0, len(log_snapshot) - 1, 2)` → `range(0, len(log_snapshot), 2)` with a bounds guard on `i + 1`.

---

### `tests/test_file_watcher.py` — dead variable `lock` in `test_session_lock_is_acquired_during_import`

`lock = threading.Lock()` is created but never referenced again — `InstrumentedLock` constructs its own internal lock. The named variable is disconnected from the instrumentation and will confuse future readers.

---

### `tests/test_file_watcher.py` — missing teardown in `test_session_lock_is_acquired_during_import`

If `assert held.wait(timeout=5.0)` fails, the import thread stays blocked in `InstrumentedLock.acquire()` waiting on `release_gate`. The `clean_watcher` autouse fixture eventually calls `stop_watching()` → `obs.join(timeout=5.0)`, causing a 5-second stall that can cascade across subsequent tests. Consider setting `release_gate` in a `finally` block, or calling `fw.stop_watching()` explicitly before the failing assertion propagates.

---

jweber93:
Thanks for the thorough review. All four issues addressed:

1. **RLock type annotation** — Changed `Optional[threading.Lock]` → `Optional[threading.RLock]` on both `_ZROHandler.__init__` and `start_watching`. Added docstring warning about the re-entrancy requirement (`session_saver` re-acquires via `SessionStore.save_session`).

2. **Odd-length log IndexError** — Added `assert len(log_snapshot) % 2 == 0` before the loop and changed `range(0, len(log_snapshot) - 1, 2)` → `range(0, len(log_snapshot), 2)`.

3. **Dead variable `lock`** — Removed the unused `lock = threading.Lock()` from `test_session_lock_is_acquired_during_import`.

4. **Teardown safety** — Wrapped `held.wait()` in `try/finally` that always sets `release_gate`, preventing the import thread from blocking indefinitely on test failure.

PR #327, opened 2026-06-28
Remove superseded scaffolding: scrollToCard, header status buttons, dead CSS

### LINKED ISSUE #298: Cleanup: remove superseded scaffolding
**Files:** `App.jsx`, `index.css`  ·  **Depends on:** #2, #3, #4, #7

### Change
- Remove `ProgressBar`, `QualityDot`, `scrollToCard`, and the header status buttons that scrolled to `#bridge-card`/`#watch-folder`/`#dogegen-card` (now redundant — status lives in `PhaseRail` + `InstrumentPanel`).
- Drop dead CSS: `.progress-bar`, `.step-tab*`, `.action-card`/`.action-icon` (if fully replaced), `.card-pulse` if the scroll-pulse is gone.

### Acceptance
- No references to removed symbols remain; `npm run build` is clean.

jweber93:
Removes the `scrollToCard` function and three header status buttons that scrolled to `#bridge-card`/`#watch-folder`/`#dogegen-card` (now redundant — status lives in `PhaseRail` + `InstrumentPanel`).

Drops dead CSS: `.action-card`, `.card-pulse`, `@keyframes cardPulse`, and unused `MEASUREMENT_STEPS` const.

Also fixes a behavioral gap: `InstrumentPanel` now renders on all session steps (not just measurement steps) so bridge/dogegen/watch/ADB status is always visible, not only during measurement phases.

Closes #298

### DIFF
--- a/frontend/src/App.jsx
+++ b/frontend/src/App.jsx
@@ -48,16 +48,6 @@ import { LogsPanel }     from './components/LogsPanel';
 import { AppShell }     from './components/AppShell';
 
 
-function scrollToCard(id) {
-  const el = document.getElementById(id);
-  if (!el) return;
-  el.scrollIntoView({ behavior: 'smooth', block: 'center' });
-  el.classList.remove('card-pulse');
-  void el.offsetWidth;
-  el.classList.add('card-pulse');
-  setTimeout(() => el.classList.remove('card-pulse'), 900);
-}
-
 export default function App() {
   const sess = useSession();
   const { session, profiles, watchStatus, watchDefaultPath, bridgeStatus, bridgeUrl, dogegenStatus, adbStatus, loading,
@@ -120,11 +110,6 @@ export default function App() {
     setShowComparison(prev => !prev);
   }
 
-  const bridgeOk = bridgeStatus?.ok;
-  const bridgeConfigured = bridgeStatus?.configured;
-  const dogegenRunning = dogegenStatus?.running;
-  const dogegenConfigured = dogegenStatus?.configured;
-
   return (
     <div style={{ display: 'flex', flexDirection: 'column', minHeight: '100vh' }}>
       {showStartOver && (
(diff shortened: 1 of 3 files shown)


jweber93:
## Review: PR #327 — Remove superseded scaffolding

This PR deletes `scrollToCard` and its three caller buttons (WATCHING / Bridge / Dogegen) from the header, and removes the now-orphaned `.action-card`, `.card-pulse`, and `@keyframes cardPulse` CSS. The diff itself is clean — all four derived `const` declarations (`bridgeOk`, `bridgeConfigured`, `dogegenRunning`, `dogegenConfigured`) are properly removed in the same hunk as the JSX that references them, so there's no dangling `ReferenceError` risk.

One real behavioral gap survived verification:

---

**`frontend/src/components/AppShell.jsx:18` — Status indicators are now invisible on non-measurement steps**

`InstrumentPanel` (which now owns bridge/watch/dogegen status) is only mounted when `MEASUREMENT_STEPS.has(session.step)` is true — i.e., on `pre_grayscale`, `luminance`, `white_balance`, etc. The deleted header buttons were rendered unconditionally and were visible on every step, including Setup, Prepare, and any other non-measurement page. After the PR, if the watch folder goes active, the bridge drops offline, or dogegen starts while the user is on a non-measurement step, there is no ambient indicator anywhere — `PhaseRail` has zero reference to `bridgeStatus` or `dogegenStatus`, and `InstrumentPanel` is simply not in the DOM.

**Scenario:** User is on the Prepare step and their bridge service goes offline. Previously: a red "Bridge" button appeared in the header immediately. After this PR: no indication until they advance to a measurement step and InstrumentPanel renders.

---

Everything else checked out:
- All JSX references to deleted variables are fully covered by the same diff hunks — no orphaned identifiers.
- `.action-card` had zero JSX consumers before the PR, so its removal is safe.
- The `scrollToCard('#watch-folder')` and `scrollToCard('#bridge-card')` targets didn't exist in the DOM (those IDs were never set), so those two buttons were already silently broken — removal is correct.
- `InstrumentPanel` unconditionally renders the ZRO Bridge row (always shows connected/offline) so the `bridgeConfigured` vs `bridgeOk` distinction isn't a gap when InstrumentPanel is actually visible.

---

PR #330, opened 2026-06-28
compute history metrics on-the-fly when missing from entries (#268)

### LINKED ISSUE #268: [Bug] History/comparison endpoint returns None for all computed metrics
## Summary

`server.py` `get_report_history()` endpoint (line 1503) reads computed fields like `post_grayscale_avg_de`, `gamma_avg`, `wb_avg_de`, `cms_avg_de` from serialized session entries. However, `serialize_session()` in `calibrator/session.py` does NOT include these computed fields — they are only computed on-the-fly by `report_payload()`. The history endpoint always returns `None` for all metric fields.

## Evidence

```python
# server.py lines 1519-1529
sessions.append({
    "session_id": baseline.get("session_id"),
    "date": baseline.get("date"),
    "mode": baseline.get("mode"),
    "is_baseline": True,
    "avg_de": baseline.get("post_grayscale_avg_de"),  # Always None — key doesn't exist in serialized session
    "peak_luminance": baseline.get("peak_luminance"),
    "gamma_avg": baseline.get("gamma_avg"),  # Always None
    "wb_avg_de": baseline.get("wb_avg_de"),  # Always None
    "cms_avg_de": baseline.get("cms_avg_de"),  # Always None
})
```

The same pattern applies to non-baseline history entries (lines 1531-1542).

## Impact

The comparison page shows all metrics as "—" for every session. The comparison feature is effectively broken for viewing computed metrics. Users cannot compare calibration quality across sessions.

## Suggested Fix

Either:
1. Compute these fields in `_load_history()` by calling `report_payload()` or a shared helper per session.
2. Store computed stats in the serialized session when the report is generated (preferred — also improves API performance).

**Priority:** High  
**Component:** server / history API  
**Files:** `server.py` lines 1503-1544, `calibrator/session.py` (`serialize_session`)

jweber93:
## Summary

Fixes the history/comparison endpoint returning `None` for all computed metrics.

### Problem

The `/api/report/history/{tv_key}` endpoint reads serialized session entries that may not have computed metric fields (`post_grayscale_avg_de`, `gamma_avg`, `wb_avg_de`, `cms_avg_de`). These fields are only computed on-the-fly by `report_payload()`.

### Solution

Add helper functions to compute metrics from the session's full report payload when history entries are missing them:
- `_compute_metrics_from_session(session_id)` — loads session and computes metrics from report_payload()
- `_history_entry_metrics(entry)` — returns stored values if available, otherwise computes on-the-fly

### Changes

- `server.py`: Added metric computation helpers and updated history endpoint to use them
- `tests/test_history_metrics.py`: Added comprehensive tests for the fix

### Testing

All 5 new tests pass:
- History entries without metrics get computed on-the-fly  
- Baseline entries without metrics get computed on-the-fly
- Entries with existing metrics are not recomputed
- Missing sessions are handled gracefully
- Computed metrics match report_payload values

Closes #268

### DIFF
--- a/calibrator/history.py
+++ b/calibrator/history.py
@@ -126,6 +126,85 @@ def load_baseline(tv_key: str) -> Optional[Dict[str, Any]]:
         return None
 
 
+def update_history_entry(tv_key: str, session_id: str, metrics: Dict[str, Any]) -> bool:
+    """Update an existing history entry with computed metrics.
+
+    Args:
+        tv_key: TV profile key
+        session_id: Session identifier to update
+        metrics: Dict of metric keys and values to merge into the entry
+
+    Returns:
+        True if entry was found and updated, False otherwise.
+    """
+    path = _sessions_path(tv_key)
+    if not path.exists():
+        return False
+
+    entries: List[Dict[str, Any]] = []
+    try:
+        with open(path, encoding="utf-8") as fh:
+            for line in fh:
+                line = line.strip()
+                if line:
+                    try:
+                        entries.append(json.loads(line))
+                    except json.JSONDecodeError:
+                        pass
+    except OSError:
+        return False
+
+    updated = False
+    for entry in entries:
+        if entry.get("session_id") == session_id:
+            entry.update(metrics)
+            updated = True
+
+    if not updated:
+        return False
+
+    # Rewrite the entire file with updated entry
+    try:
+        path.parent.mkdir(parents=True, exist_ok=True)
+        with open(path, "w", encoding="utf-8") as fh:
+            for entry in entries:
+                fh.write(json.dumps(entry) + "\n")
+    except OSError:
+        return False
+
+    return True
+
+
+def update_baseline(tv_key: str, metrics: Dict[str, Any]) -> bool:
+    """Update the baseline entry with computed metrics.
+
+    Args:
+        tv_key: TV profile key
+        metrics: Dict of metric keys and values to merge into the baseline
+
+    Returns:
+        True if baseline was found and updated, False otherwise.
+    """
+    path = _baseline_path(tv_key)
+    if not path.exists():
+        return False
+
+    try:
+        baseline = json.loads(path.read_text(encoding="utf-8"))
+    except (json.JSONDecodeError, OSError):
+        return False
+
+    baseline.update(metrics)
+
+    try:
+        with open(path, "w", encoding="utf-8") as fh:
+            json.dump(baseline, fh, indent=2)
+    except OSError:
+        return False
+
+    return True
+
+
 def history_summary(tv_key: str) -> Dict[str, Any]:
     """Return a lightweight summary dict for API exposure.
 
--- a/server.py
+++ b/server.py
@@ -61,6 +61,8 @@ from calibrator.history import (
     load_baseline as _load_baseline,
     load_history as _load_history,
     record_session as _record_session,
+    update_baseline as _update_baseline,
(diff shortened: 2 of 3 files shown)


jweber93 (commented):
## Review

This PR fixes the root symptom correctly — history entries that lack stored metrics now return computed values instead of `None`. The helper decomposition is clean. A few issues worth addressing before merging, ranging from a latent performance problem to a missing `logger.warning`.

---

### 🔴 N+1 blocking disk reads per request

`_history_entry_metrics` is called once per history entry (up to 20 + baseline). Each call that finds any `None` metric triggers `_compute_metrics_from_session`, which does:
1. A synchronous `SESSION_STORE_DIR/{sid}.json` read via `_load_session_by_id`
2. A full `_report_payload` computation (all calibration metrics)

For a user whose older entries predate metric storage, a single `GET /api/report/history/{tv_key}` issues 21 sequential blocking reads + report computations on the FastAPI thread. This is a significant regression for a listing endpoint that was previously O(1) dict lookups.

**Bigger problem:** the computed values are returned in-memory but never written back to the history store, so every subsequent page load for that TV repeats the full cost indefinitely.

Two options, either is fine:
- **Write-back on compute**: persist the backfilled metrics back to the history entry after computing them, so the N+1 cost is paid once.
- **Fix at the write path**: in `_record_session`, store all 5 metric keys from the `report` dict that's already available there. Old entries would still need backfill on first read, but after that they're persisted.

---

### 🟡 `except Exception` swallows errors silently — no logging

```python
# server.py ~line 1519
except Exception:
    return {
        "post_grayscale_avg_de": None,
        ...
    }
```

If `_load_session_by_id` raises `HTTPException(500, ...)` (corrupt session file) or `_report_payload` throws for any other reason, the error disappears completely. The UI chart renders nulls with no indication anything went wrong, and nothing lands in the logs.

At minimum:
```python
except Exception:
    logger.warning("Failed to compute metrics for session %s", session_id, exc_info=True)
    return {"post_grayscale_avg_de": None, ...}
```

---

### 🟡 `any(v is None)` trigger includes `peak_luminance`, causing unnecessary recomputes

```python
# server.py ~line 1549
if any(v is None for v in metrics.values()):
```

`peak_luminance` is included in the check. For sessions where luminance was never measured, `peak_luminance` is legitimately `None` — but this condition will still fire, pay the full disk read + `_report_payload` cost, and then return `None` for `peak_luminance` anyway. The trigger should only fire for the four *computed* metrics (`post_grayscale_avg_de`, `gamma_avg`, `wb_avg_de`, `cms_avg_de`), not for `peak_luminance` which is stored directly:

```python
COMPUTED_KEYS = {"post_grayscale_avg_de", "gamma_avg", "wb_avg_de", "cms_avg_de"}
if any(metrics.get(k) is None for k in COMPUTED_KEYS):
    ...
```

---

### 🔵 Redundant `finally: pass` and manual monkeypatch restores in tests

Several tests have empty `finally: pass` blocks that do nothing — looks like copy-paste leftovers. Also, `test_history_entry_without_metrics_is_computed` manually restores `_load_history` in a `finally` block via `monkeypatch.setattr`, which is unnecessary since `monkeypatch` already auto-restores everything at test teardown. These are minor but add noise.

---

**Summary**: The correctness fix is solid. The performance issue (N+1 reads with no write-back) is the main thing worth resolving before this goes to production, since it can make the history page slow for any TV with legacy entries — and that slowness repeats on every page load.

---

PR #333, opened 2026-06-28
Hold SessionStore RLock across import get-mutate-save (#256)

### LINKED ISSUE #256: File watcher auto-import mutates session without lock
## Root Cause

In `calibrator/file_watcher.py`, the `_ZROHandler._import_file()` method mutates session state **without holding any lock**.

## Evidence

Lines 462-518 perform session mutations unprotected:

```python
def _import_file(self, src_path: str) -> None:
    session = self._session_getter()  # Returns live session dict
    # ... no lock acquired here ...
    
    # Lines 468-518: Direct mutations happen unsafely
    for key, meas_dicts in bucket_map.items():
        session[key].append(measurement)  # Unprotected write!
    
    session["peak_luminance"] = round(latest_y, 2)  # Unprotected write!
    self._session_saver()  # Writes to disk while possibly being read
```

The class has `_lock` (line 239) but it only protects status fields like `_watcher_error`, not the actual session data.

## Impact

- Concurrent file watcher imports + manual CSV uploads = corrupted session
- Peak luminance value can be overwritten mid-calculation
- Measurement deduplication logic can fail with duplicate entries

## Fix Strategy

Option A: Lock in file watcher
```python
def _import_file(self, src_path: str) -> None:
    with _lock:  # Acquire global lock
        session = self._session_getter()
        # ... all mutations here ...
```

Option B: Pass lock from server.py (cleaner)
- SessionStore should expose its lock
- File watcher acquires that lock before mutating

## Acceptance Criteria

- [ ] Session mutations in `_import_file()` are protected by a lock
- [ ] Lock scope covers both read (getter) and write (saver) operations
- [ ] Test concurrent auto-import + manual upload scenarios
- [ ] Verify no deadlocks with SessionStore operations

jweber93:
Closes #256

## Problem

`SessionStore.import_zro_bytes` and `import_generic_bytes` called `get()` and `save_session()` under separate lock acquisitions, leaving session dict mutations unprotected between the two. Concurrent file-watcher auto-imports and manual CSV uploads could interleave, causing:

- Corrupted measurement lists (duplicate/lost entries)
- peak_luminance overwritten mid-calculation
- Measurement deduplication logic failures

The file watcher side was already partially fixed (`_do_import_file` holds `session_lock`), but the `SessionStore` import methods had the same gap.

## Fix

Move CSV parsing outside the lock (CPU-bound, no shared state), then wrap the full `get() → mutate → save_session()` sequence in a single `with self._lock:` block. The RLock is reentrant so `save_session`'s own acquisition is safe — no deadlock risk.

## Tests

Added `TestFileWatcherAndManualUpload` (4 tests):
- File-watcher + `import_zro_bytes` serialise via RLock
- Lock is held across the full get-mutate-save span
- `peak_luminance` not corrupted under concurrent imports
- No deadlock on reentrant lock

Updated existing `test_simultaneous_import_zro_bytes_no_data_loss` and `test_concurrent_import_generic_bytes` to remove 'known race' caveats.

All 1083 tests pass.

### DIFF
--- a/calibrator/session.py
+++ b/calibrator/session.py
@@ -2039,34 +2039,34 @@ class SessionStore:
         self.save_session(sid)
         return session
 
-    def import_zro_bytes(
+    def _locked_import(
         self,
         sid: str,
         filename: Optional[str],
-        contents: bytes,
+        bucket_map: Dict[str, List[dict]],
+        step_warnings: List[str],
+        total_rows: int,
+        sessions_detected: int,
+        colour_sessions: int,
+        grayscale_sessions: int,
+        session_breaks: int,
+        unknown_rows: int,
+        raw_rows: List,
     ) -> Tuple[Dict[str, Any], Dict[str, Any]]:
-        try:
-            result = parse_zro_csv(contents)
-        except Exception as exc:
-            raise HTTPException(400, f"ZRO CSV parse error: {exc}") from exc
+        """Merge *bucket_map* into the session atomically under the store RLock.
+
+        The RLock is reentrant (``threading.RLock``) so that
+        ``save_session``'s own ``with self._lock:`` re-enters without
+        deadlock.  This serialises the full get → mutate → save sequence
+        against concurrent file-watcher auto-imports and other callers
+        that share the same lock (``store._lock``).
+        """
         with self._lock:
             session = self.get(sid)
             if session.get("mode") is None:
                 raise HTTPException(
                     400, "Select a calibration mode before importing measurements."
                 )
-            result = recontextualize_import_result(
-                result,
-                session.get("signal_range", "full"),
-                session.get("code_scale", "8bit"),
-            )
-            if result.total_rows == 0:
-                raise HTTPException(
-                    400,
-                    "No valid measurement rows found. Expected a tab-separated ZRO export with columns: Date and time, R, G, B, Y, x, y, msec.",
-                )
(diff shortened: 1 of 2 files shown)


jweber93:
✅ Review: Hold SessionStore RLock across import get-mutate-save (#256)
Status: Ready to merge (low risk)
Tests: All 1083 passing ✅

Core Fix
CSV parsing now happens outside the lock (CPU-bound), then the full get() → mutate → save_session() sequence is atomic within with self._lock:. RLock is reentrant, so save_session()'s re-acquisition is safe—no deadlock risk.

Test Coverage
Excellent—4 new tests cover:

File-watcher + manual upload serialization via RLock
Lock held across full get-mutate-save span (instrumented test)
peak_luminance not corrupted under concurrent imports
No deadlock on reentrant lock acquisition
Updated test docstrings properly remove "known race" caveats.

Suggestions
Consider a helper method for the locked sequence — Both import_zro_bytes and import_generic_bytes now have identical lock-wrapping patterns. A utility like _locked_import(sid, mutator_fn) could reduce duplication and ensure future imports are automatically correct. (Nice-to-have refactoring, not a blocker.)

Add inline comment explaining RLock choice — Brief note near with self._lock: explaining why RLock is used and why re-acquisition in save_session() is safe. Helps future maintainers.

Audit other mutation patterns — Quick check: are there other get() → mutate → save_session() sequences elsewhere in SessionStore that need the same fix? Flag any as follow-ups.

PR #365, opened 2026-06-29
fix: normalize CalibrationTarget.primaries to dict format (#362)

### LINKED ISSUE #362: [BUG] CMS gamut mapping crashes with AttributeError when target.primaries is a tuple
## Summary

`calibrator/reports.py:target_xy_for_colour()` assumes `target.primaries` is a dict and calls `.items()`. When the frontend sends primaries as a tuple (the JSON format from `calcore/target.py`), this crashes with:

```
AttributeError: 'tuple' object has no attribute 'items'
```

## Reproduction

1. Create a calibration target with `primaries` as a tuple:
```python
target = CalibrationTarget(
    gamut="bt709",
    eotf="bt1886",
    primaries=((0.64, 0.33), (0.30, 0.60), (0.15, 0.06)),
)
```

2. Take a measurement with `cms=True` and label like `"Red"` or `"Green"`.

3. Call `report_payload(session)` — it crashes in `target_xy_for_colour()` at line:
```python
for color_name, (px, py) in target.primaries.items():  # tuple has no .items()
```

## Root Cause

`calcore/target.py:CalibrationTarget.__init__()` accepts `primaries` as either a dict or tuple of tuples. When it's a tuple, the comparison report path in `calibrator/reports.py` doesn't handle it.

## Fix Options

**Option A (recommended):** Convert primaries to dict in `CalibrationTarget.__init__()` so downstream code always sees a dict:
```python
if isinstance(primaries, tuple):
    self.primaries = {name: coords for name, coords in zip(["red", "green", "blue"], primaries)}
else:
    self.primaries = dict(primaries)
```

**Option B:** Add a defensive check in `target_xy_for_colour()`:
```python
if isinstance(target.primaries, tuple):
    # Convert to dict format expected by this function
    color_names = ["red", "green", "blue"]
    primaries_dict = dict(zip(color_names, target.primaries))
else:
    primaries_dict = target.primaries
```

## Impact

Any user who takes CMS measurements (e.g., using a colorimeter with the app's built-in targets) will get a crash when viewing the comparison report. The grayscale-only path works fine.

jweber93:
## 📺 Overview

Fixes #362 - CMS gamut mapping crashes with AttributeError when `target.primaries` is a tuple.

### What does this PR do?
* **Type of change:** Bug fix
* **Component(s) affected:** Calibration target data model, CMS/color tuner measurements

---

* **The Problem:** `CalibrationTarget` accepts `primaries` as either a dict or tuple of tuples (the JSON serialization format from the frontend). However, functions like `target_xy_for_colour()` and `target_nits_for_colour()` in `calibrator/guidance.py` assume `target.primaries` is always a dict and access it with keys like `primaries["red"]`. When primaries is passed as a tuple, this raises `AttributeError: 'tuple' object has no attribute 'items'`.

* **The Solution:** Normalize `primaries` to dict format in `CalibrationTarget.__post_init__()`. If primaries is a tuple, convert it to a dict with keys ["red", "green", "blue"] in order. If it's already a dict (or dict-like), convert it to a fresh dict to ensure consistency.

* **Impact on existing workflows/APIs:** None. This is a transparent fix that ensures downstream code always receives primaries as a dict, which is what it expects. Existing code that passes primaries as a dict will continue to work identically.

---

### Verification Steps
1. CalibrationTarget can now be created with primaries as a tuple:
   ```python
   target = CalibrationTarget(
       mode=CalMode.SDR,
       gamut="bt709",
       eotf="BT.1886",
       primaries=((0.64, 0.33), (0.30, 0.60), (0.15, 0.06)),
   )
   ```
2. The primaries field is automatically converted to dict format with keys "red", "green", "blue"
3. Functions like `target_xy_for_colour()` now work correctly regardless of input format

---

## 🧼 Checklist
* [x] Code adheres to project style guidelines
* [x] Fix verified with manual testing of both tuple and dict formats
* [x] No breaking changes to existing API

---
_Generated by [Claude Code](https://claude.ai/code/session_01FpbWucnk4ywPbNRkJjjmkj)_

### DIFF
--- a/calcore/models.py
+++ b/calcore/models.py
@@ -246,7 +246,7 @@ class CalibrationTarget:
         # Normalize primaries to dict format if it's a tuple (JSON serialization format).
         if isinstance(self.primaries, tuple):
             color_names = ["red", "green", "blue"]
-            self.primaries = {name: coords for name, coords in zip(color_names, self.primaries)}
+            self.primaries = dict(zip(color_names, self.primaries))
         else:
             self.primaries = dict(self.primaries)
         # Normalize display names that older calibrator code expects.
--- a/calcore/models.py
+++ b/calcore/models.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from collections.abc import Mapping, Sequence
 from dataclasses import dataclass, field
 from enum import Enum
 import math
@@ -243,12 +244,41 @@ class CalibrationTarget:
     def __post_init__(self) -> None:
         if not self.primaries:
             self.primaries = detect_primaries(self.gamut)
-        # Normalize primaries to dict format if it's a tuple (JSON serialization format).
-        if isinstance(self.primaries, tuple):
-            color_names = ["red", "green", "blue"]
-            self.primaries = dict(zip(color_names, self.primaries))
-        else:
+
+        # Normalize primaries to dict format {name: (x, y)}.
+        # Handles: dict (pass-through), tuple/list of coordinate pairs (JSON format).
+        if isinstance(self.primaries, Mapping):
             self.primaries = dict(self.primaries)
+        elif isinstance(self.primaries, Sequence) and not isinstance(self.primaries, (str, bytes)):
+            # Validate shape: sequence of 3 coordinate pairs (x, y)
+            if len(self.primaries) != 3:
+                raise ValueError(
+                    f"primaries sequence must have exactly 3 elements (red, green, blue), got {len(self.primaries)}"
+                )
+            color_names = ["red", "green", "blue"]
+            validated_primaries = {}
+            for i, (name, coords) in enumerate(zip(color_names, self.primaries)):
+                if not isinstance(coords, Sequence) or isinstance(coords, (str, bytes)):
+                    raise ValueError(
+                        f"primaries[{i}] ({name}) must be a sequence of (x, y) coordinates, got {type(coords).__name__}"
+                    )
+                if len(coords) != 2:
+                    raise ValueError(
+                        f"primaries[{i}] ({name}) must be a pair (x, y), got {len(coords)} elements"
+                    )
+                try:
+                    x, y = float(coords[0]), float(coords[1])
+                    validated_primaries[name] = (x, y)
+                except (TypeError, ValueError) as e:
+                    raise ValueError(
+                        f"primaries[{i}] ({name}) coordinates must be numbers (float/int), got {type(coords[0]).__name__}, {type(coords[1]).__name__}"
+                    ) from e
+            self.primaries = validated_primaries
+        else:
+            raise TypeError(
+                f"primaries must be a mapping or sequence of (x, y) pairs, got {type(self.primaries).__name__}"
+            )
+
         # Normalize display names that older calibrator code expects.
         if self.gamut.lower() in ("bt709", "709", "rec709"):
             self.gamut = "Rec.709"
--- a/calcore/models.py
+++ b/calcore/models.py
@@ -245,7 +245,8 @@ class CalibrationTarget:
             self.primaries = detect_primaries(self.gamut)
         # Normalize primaries to dict format if it's a tuple (JSON serialization format).
         if isinstance(self.primaries, tuple):
-            self.primaries = {name: coords for name, coords in zip(["red", "green", "blue"], self.primaries)}
+            color_names = ["red", "green", "blue"]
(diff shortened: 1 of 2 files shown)


jweber93:
Notes and suggested improvements (actionable)

Make type / shape detection more robust

Problem: the code only checks isinstance(..., tuple). The frontend or other code may send lists (list-of-tuples) or other sequence types. Also dict(self.primaries) may silently behave incorrectly if self.primaries is a sequence of coordinate tuples rather than (key, value) pairs.
Suggestion: explicitly detect mapping vs sequence vs iterable-of-coordinate-pairs:
Use collections.abc.Mapping to detect mapping-like inputs and copy: if isinstance(self.primaries, Mapping): self.primaries = dict(self.primaries)
Else if it's a Sequence (tuple or list) of length 3 where each element is a 2-tuple of (x,y): map to ["red","green","blue"] with validation.
Else raise a clear ValueError with an informative message (or log and fallback).
This prevents accidentally calling dict() on the wrong shape and getting confusing exceptions later.
Validate contents and lengths

Add checks: sequence length == 3, each element is a pair (len==2) of numbers (ints/floats).
If validation fails, raise an explicit error (TypeError/ValueError) or fall back to detect_primaries(self.gamut) depending on desired behavior.
This makes failures earlier and clearer and avoids downstream cryptic errors.
Accept lists and other JSON-produced shapes

Replace isinstance(self.primaries, tuple) with a check that covers both tuple and list, or better yet, Sequence detection as above. JSON will more commonly deserialize to lists.
Avoid silently converting arbitrary iterables with dict()

The fallback else: self.primaries = dict(self.primaries) can raise or produce unexpected mapping if the input is e.g. [(0.64, 0.33), ...] (it will treat each coordinate tuple as a key/value pair, which is wrong). Use explicit branches so only well-shaped inputs are converted.
Add unit tests

Add tests that cover:
primaries as dict (existing behavior)
primaries as tuple-of-tuples
primaries as list-of-lists
malformed primaries (wrong length, wrong element shape) and assert that an informative error is raised (or fallback behavior).
Tests make this non-regressive and document expected behavior.
Docstring / inline comment & typing

Add a short docstring or comment on CalibrationTarget.primaries format expected (e.g., Mapping[str, Tuple[float, float]]).
If project uses typing, annotate the field type and update post_init guard to match.
PR status: mark ready for review

The PR is currently a draft (draft: true). When you want reviewers to review, mark it ready for review.
Merge-readiness and risk assessment

Readiness: Almost ready; the change is small and targeted. I recommend:
Add the more robust normalization/validation (see above),
Add unit tests,
Mark PR as ready for review (remove draft) after tests are added.
Risk level: Low if validation is added; Medium if left as-is because of potential unhandled input shapes.
Concrete suggested code sketch (for post_init)

In prose: first check Mapping, then check Sequence with length 3 and shape validation, else raise ValueError. Example (conceptual):
if isinstance(self.primaries, Mapping): self.primaries = dict(self.primaries)
elif isinstance(self.primaries, Sequence) and not isinstance(self.primaries, (str, bytes)):
if len(self.primaries) == 3 and all(len(p)==2 and is_number for p in self.primaries): map to red/green/blue
else: raise ValueError("primaries must be mapping or sequence of three (x,y) pairs")
else: raise TypeError("Unsupported primaries format...")

jweber93:
Thank you for the thorough review! I've implemented all the suggested improvements:

## Changes Made

✅ **Robust type detection:**
- Use `collections.abc.Mapping` to detect dict-like inputs
- Use `collections.abc.Sequence` to detect tuple/list inputs
- Reject unsupported types with informative `TypeError`

✅ **Comprehensive validation:**
- Validate sequence length (exactly 3 elements for red/green/blue)
- Validate each element is a coordinate pair (length 2)
- Validate coordinates are numeric (int or float)
- Raise clear `ValueError` with context (index, color name, issue)

✅ **Handle JSON deserialization formats:**
- Dict (existing behavior)
- Tuple of tuples
- List of lists
- List of tuples
- Integer coordinates are converted to float

✅ **Unit tests added:**
- Test all input formats work correctly
- Test error cases raise appropriate exceptions
- Test integer → float conversion
- Test edge cases (wrong length, non-numeric, unsupported types)

✅ **Code clarity:**
- Added inline comments explaining each validation step
- Error messages include the specific coordinate index and color name

## Test Results
All manual tests pass (7/7) covering valid inputs and all error cases.

The PR is now ready for review. CI should re-run with the new test suite.

---

PR #388, opened 2026-06-29
Fix: Use correct gamma (2.2) for SDR BT.1886 EOTF

### LINKED ISSUE #366: Bug: EOTF used as gamma in grayscale target computation for SDR sessions (calcore/targets.py:36)
## Problem

In `calcore/targets.py` line 36, the BT.1886 EOTF is computed with a **hardcoded gamma of 2.4**:

```python
target_y = bt1886_eotf(n, measured_peak_y, measured_black_y, gamma=2.4)
```

The SDR calibration target in `calcore/models.py:256` defines `gamma=2.2`. The reference luminance used for delta-E and gamma tracking should use the target's actual gamma (2.2), not a hardcoded 2.4.

## Root Cause

The `gamma` parameter is passed directly to `bt1886_eotf()`, which uses it as the reference gamma exponent. This means all SDR grayscale target Y values are computed with gamma=2.4 instead of the configured 2.2, shifting every reference luminance downward by ~10% in midtones.

## Impact

- All SDR grayscale delta-E values are inflated because the target Y is too low
- Gamma tracking recommendations will suggest incorrect corrections (e.g., telling user to darken when they should brighten, or vice versa)
- The error compounds: a 2.4 vs 2.2 gamma difference at 50% stimulus is ~8% luminance error

## Fix Strategy

Pass the target's gamma to `target_xyz_for_patch()` and use it in the BT.1886 EOTF call:

```python
# In calcore/targets.py, the function signature should accept gamma:
def target_xyz_for_patch(patch, code_max, cfg, measured_peak_y, measured_black_y, gamma: float = 2.4):
    if patch.is_grayscale:
        ...
        elif cfg.eotf.lower() == "bt1886":
            target_y = bt1886_eotf(n, measured_peak_y, measured_black_y, gamma=gamma)
```

The caller in `calcore/analysis.py` should derive gamma from the session's target:

```python
gamma = session.get("target")?.gamma if session.get("mode") == "SDR" else 2.4
```

## Labels

bug, high-priority, math-error

jweber93:
## Summary
Fixes #366: The BT.1886 EOTF calculation was hardcoded to use gamma=2.4, but the SDR target specifies gamma=2.2. This caused ~10% luminance error in midtones, inflating delta-E values and causing incorrect gamma tracking recommendations.

## Root Cause
In `calcore/targets.py:36`, the BT.1886 EOTF was computed with a hardcoded gamma of 2.4, regardless of the calibration mode. For SDR sessions, the correct gamma is 2.2 (as defined in `SDR_TARGET`).

## Changes
- `calcore/targets.py`: Modified the BT.1886 case to derive gamma from the calibration mode
  - SDR mode: gamma=2.2 (matching SDR_TARGET definition)
  - Other modes: gamma=2.4 (default)
- Added inline comment explaining the mode-based gamma derivation for future maintainers

### Unit Tests
- All existing unit tests pass (43 passed)
- Fix verified with gamma_midtones remaining in plausible range (1.5–3.5)

### Golden Dataset Regression
⚠️ **Expected baseline changes**: Grayscale delta-E metrics increased because reference Y values now use correct gamma (2.2):
- `grayscale_avg_de`: 1.19 → 2.34 (+96%)
- `grayscale_max_de`: 2.09 → 4.24 (+103%)
- `grayscale_over_3`: 0 → 5 patches

This is **correct behavior**. The reference dataset's measured gamma (~2.35) is now further from the corrected target (2.2), showing that the display needs more aggressive gamma adjustment. The old baseline was artificially low due to the buggy gamma=2.4 target.

### Downstream Impact Assessment
- **No breaking changes**: Higher delta-E values flow through existing decision logic correctly
  - LLM decision rules (calcore/llm.py) use thresholds of avg ΔE > 2.0 and max ΔE > 3.0 — these will now trigger appropriately for SDR sessions
  - Guidance logic (calibrator/guidance.py) checks de > 3.0 for luminance error flagging — this is now correct
- Case sensitivity verified: `cfg.mode.lower() == "sdr"` is defensive against different case inputs

---
_Generated with [Claude Code](https://claude.ai/code/session_01RN2jGsQpeGg1FiR3bA2U7h)_

### DIFF
--- a/calcore/targets.py
+++ b/calcore/targets.py
@@ -33,7 +33,8 @@ def target_xyz_for_patch(
             rel = pq_eotf(n) / 10000.0
             target_y = measured_peak_y * rel
         elif cfg.eotf.lower() == "bt1886":
-            target_y = bt1886_eotf(n, measured_peak_y, measured_black_y, gamma=2.4)
+            gamma = 2.2 if cfg.mode.lower() == "sdr" else 2.4
+            target_y = bt1886_eotf(n, measured_peak_y, measured_black_y, gamma=gamma)
         else:
             gamma = (
                 2.2
--- a/calcore/targets.py
+++ b/calcore/targets.py
@@ -33,6 +33,7 @@ def target_xyz_for_patch(
             rel = pq_eotf(n) / 10000.0
             target_y = measured_peak_y * rel
         elif cfg.eotf.lower() == "bt1886":
+            # Use target's gamma: 2.2 for SDR (per SDR_TARGET in models.py), 2.4 for others
             gamma = 2.2 if cfg.mode.lower() == "sdr" else 2.4
             target_y = bt1886_eotf(n, measured_peak_y, measured_black_y, gamma=gamma)
         else:
--- a/tests/golden/data/sdr_bt709_grayscale.json
+++ b/tests/golden/data/sdr_bt709_grayscale.json
@@ -1,7 +1,7 @@
 {
-  "grayscale_avg_de": 1.1915755841757187,
-  "grayscale_max_de": 2.088598553075223,
-  "grayscale_over_3": 0,
+  "grayscale_avg_de": 2.344320097258325,
+  "grayscale_max_de": 4.238467727676889,
+  "grayscale_over_3": 5,
   "gamma_midtones": 2.3644412093905407,
   "pq_err_midtones": null,
   "color_75_avg_de": null,


jweber93:
Constructive Feedback
Strengths:

Root cause correctly identified — The bug report in #366 is precise and reproducible
Minimal, focused change — Only the necessary lines modified; no scope creep
Mode-aware fix — Conditional logic handles both SDR (2.2) and non-SDR (2.4) paths
Test delta is valid — Baseline update shows higher delta-E, which confirms the fix is working (measured data is now compared against the correct target)
Suggestions for improvement:

Add an inline comment in the code — The mode-based gamma derivation is clever but terse. Consider:

Python
# Use target's gamma for EOTF: 2.2 for SDR (per SDR_TARGET), 2.4 for others
gamma = 2.2 if cfg.mode.lower() == "sdr" else 2.4
This makes the "why" clear to future maintainers.

Consider case sensitivity — The code checks cfg.mode.lower() == "sdr" (defensive), which is good. However, verify that AnalysisConfig.mode is always set (line 41 shows mode: str = "hdr" as default). If SDR is ever passed differently (e.g., "SDR", "Sdr"), this remains safe. ✅

Document the expected baseline changes — In the PR description, it would help reviewers if you added:

"Expected test outcome: Grayscale delta-E metrics increase (~96% avg, ~103% max) because the reference Y values now use the correct gamma (2.2). Higher delta-E is correct—it shows the measured display's actual gamma (~2.35) is now farther from the corrected target."

This prevents confusion when reviewers see "tests updated" and assume values going up is bad.

Verify downstream impact — Does any analysis or reporting code interpret the new higher delta-E values differently? For example:

Does gamma tracking use a ΔE > 3 threshold that now triggers on 5 patches instead of 0?
Does any recommendation logic that was calibrated on the old (incorrect) values need adjustment?
A brief statement of "no downstream logic changes needed" or a link to a related PR would be reassuring.

Issue #369, opened 2026-06-29
Bug: Grayscale target always uses D65 white point regardless of session target (targets.py:44)


jweber93:
## Problem

`calcore/targets.py:30-44` converts ALL grayscale patches to D65 white point (hardcoded `D65_xy`) regardless of the session's target white point.

```python
if patch.is_grayscale:
    ...
    return xyY_to_xyz(D65_xy[0], D65_xy[1], target_y)
```

Meanwhile, `calibrator/session.py:976` and `calibrator/guidance.py:339` compute reference Y using the session's target gamma (2.2 for SDR). But the grayscale targets in analysis.py are always D65, even if a session specifies a different white point.

## Root Cause

The grayscale target computation in `calcore/targets.py` does not accept or use a white point parameter. It always uses D65 (0.3127, 0.3290).

## Impact

- If a session targets a non-D65 white point (e.g., D50 for cinematic calibration), grayscale delta-E values will be inflated
- The white point mismatch is invisible in the analysis output — users see "good" delta-E but the white balance step will try to correct for errors that are actually target mismatches

## Fix Strategy

1. Add a `white_point` parameter to `target_xyz_for_patch()` (defaulting to D65 for backward compatibility)
2. Pass the session's target white point from analysis.py → targets.py
3. Update callers in `calcore/analysis.py:68-71` to extract and pass the white point

## Labels
bug, high-priority, math-error

PR #485, opened 2026-06-30
Fix: Allow grayscale targets to use session white point instead of hardcoded D65

jweber93:
## Summary

- **Issue #369**: Grayscale targets were always using D65 white point regardless of session target settings, causing delta-E inflation for non-D65 white points

## Changes

- Add `white_point_xy` parameter to `target_xyz_for_patch()` with D65 as default for backward compatibility
- Add `white_point_xy` parameter to `analyze()` function
- Update all callers in `server.py` to extract and pass the target's white point when available
- Prevents incorrect delta-E calculations when sessions target non-D65 white points (e.g., D50 for cinematic calibration)

## Test Plan

- Golden regression tests should verify correct delta-E calculations with different white points
- Backward compatibility maintained: all analyze() calls default to D65 if white_point_xy not provided

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01Shh4YEXta25iS8HtF1EDP2

---
_Generated by [Claude Code](https://claude.ai/code/session_01Shh4YEXta25iS8HtF1EDP2)_

### DIFF
--- a/calcore/analysis.py
+++ b/calcore/analysis.py
@@ -2,9 +2,9 @@ from __future__ import annotations
 
 import math
 import statistics
-from typing import Any, Dict, List, Optional, Sequence
+from typing import Any, Dict, List, Optional, Sequence, Tuple
 
-from .colour import D65_xy, ciede2000, xyz_to_lab
+from .colour import D65_XYZ, D65_xy, ciede2000, xyY_to_xyz, xyz_to_lab
 from .eotf import pq_eotf
 from .models import AnalysisConfig, Patch, Summary
 from .targets import target_xyz_for_patch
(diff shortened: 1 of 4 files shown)


jweber93:
Merge readiness and risk assessment

Readiness: Almost ready.
Code quality: changes are small, well-scoped, and preserve backward compatibility.
CI/tests: PR mentions adding golden regression tests; I don't see test files in this change — tests should be added before merging.
Review feedback: none yet.
Risk: Low. The behavior change only affects grayscale target computation, and D65 remains the default when white point not provided.
Non-blocking housekeeping:
Add/expand unit tests for both D65 and non-D65 sessions to guard against regressions.
Add a short docstring or inline comment in analyze() and target_xyz_for_patch describing why white_point_xy is threaded through and how it interacts with xyz_to_lab (so future readers understand the intent).
Possible improvements / action items (actionable review points)

Add explicit typing for analyze white_point_xy and docstrings:
Reason: the new parameter is currently untyped in analyze(); consider annotating as Optional[Tuple[float, float]] so the signature is explicit.
How: change the signature to def analyze(..., white_point_xy: Optional[Tuple[float, float]] = None) -> Summary and update the docstring to mention defaulting to D65.
Ensure xyz_to_lab's reference white handling is consistent:
Reason: LAB conversion typically relies on a reference white. Make sure xyz_to_lab uses the same reference white that was used to compute target_xyz, or pass the reference white explicitly if xyz_to_lab supports it.
How: either confirm xyz_to_lab converts raw XYZ to LAB using the XYZ values you provide (no extra reference), or update calls to xyz_to_lab to pass an explicit white reference if the function accepts one, or add a comment documenting why current behavior is safe.
Add unit/integration tests:
Reason: This is the core bugfix; add regression tests that compute delta-E for a grayscale patch for D65 and for a different white point (e.g., D50) and assert the output changes as expected.
How: add tests that call target_xyz_for_patch and analyze with white_point_xy set to a non-D65 value, and compare computed target_xyz and resulting delta-E against a golden or computed expectation.
Small nits / polish

If calcore/targets.py uses Tuple in annotations, ensure typing.Tuple is imported at the top of that file (if not already).
Consider a tiny changelog entry in the PR body or repo CHANGELOG noting this bugfix (helps users tracking behavior changes).
The PR body references an external "Claude Code" link — that's fine, but if this is an internal project you might remove that link or add a brief human-written summary for the commit message.

PR #487, opened 2026-06-30
Fix EOTF misapplied as gamma in HDR effective-gamma calculation

### LINKED ISSUE #367: Bug: EOTF misapplied as gamma in effective gamma calculation (guidance.py:339, 399)
## Problem\n\nIn `calibrator/guidance.py`, the effective gamma calculation uses a plain gamma function instead of the proper EOTF for HDR sessions.\n\n**Line 339** (`gamma_recommendations`):\n\nThe `gamma_from_luminance()` function computes:  — this is a plain gamma exponent, not an EOTF inverse. For HDR sessions with PQ EOTF, this produces incorrect effective gamma values.\n\n**Line 399** (`u8g_gamma_control_plan`):\n\nSame issue — used for per-point gamma tracking in the U8G control plan.\n\n## Root Cause\n\nThe `gamma_from_luminance()` helper in `calibrator/utils.py:51-58` assumes a power-law EOTF:\n\nFor PQ (ST.2084), the correct inverse would use the PQ OETF, not a simple gamma.\n\n## Impact\n\n- HDR gamma tracking values are meaningless (off by 20-40% in midtones)\n- Gamma control plan recommendations for HDR mode are wrong\n- The "overall gamma is close" check at guidance.py:347/353 compares against 2.2 target, which is wrong for HDR (should use PQ tracking)\n\n## Fix Strategy\n\n1. Add a `eotf_from_luminance()` helper that handles PQ, BT.1886, and gamma EOTFs correctly\n2. Replace `gamma_from_luminance()` calls in guidance.py with the EOTF-aware version\n3. For PQ: use the inverse ST.2084 function to get normalized input, then compute gamma against that\n4. Add a note in the recommendation text that HDR gamma tracking uses PQ reference, not 2.2\n\n## Labels\n\nbug, high-priority, math-error

### LINKED ISSUE #368: Bug: EOTF misapplied as gamma in effective gamma calculation (guidance.py:339, 399)
## Problem

In `calibrator/guidance.py`, the effective gamma calculation uses a plain gamma function instead of the proper EOTF for HDR sessions.

**Line 339** (`gamma_recommendations`):
```python
eff_gamma = gamma_from_luminance(m.Y, peak_nits, stim_pct)
```

**Line 399** (`u8g_gamma_control_plan`):
```python
eff_gamma = gamma_from_luminance(m.Y, peak_nits, stim_pct)
```

The `gamma_from_luminance()` function computes: `log(Y/N) / log(stim/100)` — this is a plain gamma exponent, not an EOTF inverse. For HDR sessions with PQ EOTF, this produces incorrect effective gamma values.

## Root Cause

The `gamma_from_luminance()` helper in `calibrator/utils.py:51-58` assumes a power-law EOTF:
```python
return math.log(normalised_out) / math.log(normalised_in)
```
For PQ (ST.2084), the correct inverse would use the PQ OETF, not a simple gamma.

## Impact

- HDR gamma tracking values are meaningless (off by 20-40% in midtones)
- Gamma control plan recommendations for HDR mode are wrong
- The "overall gamma is close" check at guidance.py:347/353 compares against 2.2 target, which is wrong for HDR (should use PQ tracking)

## Fix Strategy

1. Add a `eotf_from_luminance()` helper that handles PQ, BT.1886, and gamma EOTFs correctly
2. Replace `gamma_from_luminance()` calls in guidance.py with the EOTF-aware version
3. For PQ: use the inverse ST.2084 function to get normalized input, then compute gamma against that
4. Add a note in the recommendation text that HDR gamma tracking uses PQ reference, not 2.2

jweber93:
## 📺 Overview

For PQ/HDR sessions, the effective-gamma calculation in `calibrator/guidance.py` used a plain power-law gamma instead of the proper ST.2084 EOTF, producing meaningless HDR gamma-tracking values and wrong control-plan recommendations.

Closes #367
Closes #368

### What does this PR do?
* **Component(s) affected:** Python guidance/control-plan logic (`calibrator/guidance.py`), EOTF math (`calcore/eotf.py`), shared utils (`calibrator/utils.py`)

---

* **The Problem:** `gamma_from_luminance()` computes `log(Y/peak) / log(stim/100)` — a plain power-law exponent. For PQ (ST.2084) this is not a valid EOTF inverse, so HDR gamma tracking was off by 20–40% in midtones, and the "overall gamma is close" / control-plan checks compared the result against the 2.2 SDR target (or the HDR target's gamma of 0), which is wrong for PQ.

* **The Solution:**
  * Added `pq_inverse_eotf()` (ST.2084 OETF: nits → normalised signal) in `calcore/eotf.py` — the exact inverse of `pq_eotf()`.
  * Added `eotf_from_luminance()` and `is_pq_eotf()` in `calibrator/utils.py`. For PQ it encodes the measured luminance back to a PQ signal and reports the exponent relative to the stimulus, where **1.0 == perfect PQ tracking** (>1.0 = too dark, <1.0 = too bright). Power-law / BT.1886 sessions keep the existing effective-gamma behaviour.
  * `gamma_recommendations`, `u8g_gamma_control_plan`, and `preset_gamma_control_plan` now accept the session `eotf`, compare against the PQ reference (1.0) for HDR, and surface a note that HDR tracking uses the PQ reference rather than 2.2.
  * Plumbed `target.eotf` through the `calibrator/session.py` call sites (defaulting to `"gamma"` when no target is present).

* **Impact on existing workflows/APIs:** SDR/gamma/BT.1886 behaviour is unchanged (`eotf_from_luminance` delegates to the legacy `gamma_from_luminance` for non-PQ). The three guidance functions gained a trailing optional `eotf` argument, preserving positional backward compatibility.

### Mathematical / Data Integrity Checks
* [x] PQ encode/decode validated — `pq_inverse_eotf(pq_eotf(n)) == n` to 1e-9 across the midtone range.
* [x] Floating-point precision addressed — `clamp` on the normalised PQ input; `None` guards for out-of-domain stimuli and luminance.

---

### Environment Details
* **Hardware/Mock Used:** Simulated measurements (synthetic PQ/SDR grayscale passes)

### Verification Steps
1. `python -m pytest tests/test_colour_science.py tests/test_hdr_gamma_guidance.py` — new helper + HDR guidance regression tests.
2. Full suite: `python -m pytest` — **1187 passed, 2 skipped**, coverage 80%.
3. Confirmed a perfect PQ pass reports tracking ≈ 1.0 and "close"; a half-luminance PQ pass is flagged "too dark"; SDR 2.2 passes are unchanged.

---

## 🧼 Checklist
* [x] Code adheres to project style guidelines.
* [x] Documentation (inline docstrings) updated to reflect changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0199WmKhQeLQRhNYynsruPjG

---
_Generated by [Claude Code](https://claude.ai/code/session_0199WmKhQeLQRhNYynsruPjG)_


jweber93:
Minor Suggested Improvements
Not blockers, but would increase confidence:

Extract PQ detection to a constant or enum: Instead of string matching, define:

Python
EOTF_POWER_LAW = "gamma"
EOTF_BT1886 = "BT.1886"
EOTF_PQ = "PQ (ST.2084)"

def is_pq_eotf(eotf: str) -> bool:
    return eotf in (EOTF_PQ, ...)  # Better type safety
Add a docstring example to eotf_from_luminance():

Python
"""
...
Example:
    A display measuring 500 nits at 50% stimulus under PQ:
    >>> eotf_from_luminance(500, 1000, 50, "PQ (ST.2084)")
    ~1.0  # Perfect tracking
    
    Same display but measuring 250 nits (too bright):
    >>> eotf_from_luminance(250, 1000, 50, "PQ (ST.2084)")
    ~0.5  # Below reference
"""
Add an integration test showing a real workflow (read measurement → compute guidance → check output structure). The existing tests validate the math but not the user flow. Low priority since the functions are simple wrappers.

Observations (Minor)
is_pq_eotf() robustness: String matching on "pq" and "2084" is case-insensitive and flexible, but relies on convention. If a downstream user passes "PQ-HDR" or "st.2084" (with dot), detection would fail. Consider a more explicit enum or constants. Not critical for this PR, but worth a future refactor.

Python
def is_pq_eotf(eotf: str) -> bool:
    e = (eotf or "").lower()
    return "pq" in e or "2084" in e  # ← Fragile; "st.2084" fails
Zero nits handling in pq_inverse_eotf(): Clamping to [0.0, 1.0] is correct, but the comment says "returns a tiny positive offset, not exactly 0" — this is true of the encoding, but the function actually returns ~0.0 due to clamping. Clarify:

Python
def pq_inverse_eotf(nits: float) -> float:
    # SMPTE ST 2084 forward (OETF): nits -> normalised signal in [0, 1].
    # Inverse of pq_eotf().
    # Clamping to [0, 1] ensures we stay in valid range; 0 nits encodes to ~0.
None guards in eotf_from_luminance(): Good defensive code, but the order is slightly inconsistent:

Checks stimulus_pct <= 0 or >= 100 upfront.
Then checks normalised_in <= 0 again (redundant after first check).
For PQ path, checks measured_nits <= 0 inside the if block.
Could consolidate early returns, but readability is fine as-is.

Messaging language: In gamma_recommendations(), the PQ-specific messages are good, but they assume the user understands "gamma/EOTF offset" jargon. Minor — context is clear, and this is already fairly technical output.

jweber93:
Thanks for the review. Pushed `dfdbc66` with the low-risk improvements:

- **Docstring examples** — added worked doctest examples to `eotf_from_luminance()` (perfect / too-bright / too-dark PQ tracking). Note: I used measured values, since perfect PQ tracking at 50% needs ~92 nits, not 500 — e.g. 250 nits @ 50% gives ~0.73, not ~0.5.
- **`pq_inverse_eotf()` comment** — clarified the clamping and that 0 nits encodes to a tiny positive offset (~7e-7).
- **Redundant guard** — dropped the duplicate `normalised_in <= 0` check, since `stimulus_pct` is already validated to `(0, 100)`.

A couple of the observations don't actually apply, so I left those as-is:

- **`is_pq_eotf` "fragility"** — substring matching already handles both cited cases: `is_pq_eotf("st.2084")` → `True` (contains `"2084"`) and `is_pq_eotf("PQ-HDR")` → `True` (contains `"pq"`). I kept the substring approach deliberately because it's *more* tolerant than an exact-match enum, which would break on exactly these variants. Happy to revisit with a proper `EOTF` enum in a follow-up if you'd like type safety across the codebase.
- **Integration test** — `tests/test_hdr_gamma_guidance.py` already covers the read→guidance→output flow: it builds `Measurement`s, runs them through `gamma_recommendations` / `u8g_gamma_control_plan` / `preset_gamma_control_plan`, and asserts on the recommendation text and control-plan structure.

The "gamma/EOTF offset" wording in the HDR messages is intentionally aligned with the TV's HDR tracking controls.

---

PR #496, opened 2026-06-30
fix: guard stimulus_pct null in chart axis label construction

### LINKED ISSUE #493: Bug: `m.stimulus_pct + '%'` without null guard produces literal "null%" axis labels in gamma charts
## Summary

`GammaChart.jsx` and `OverlayGammaChart.jsx` build chart axis labels via `m.stimulus_pct + '%'` with no null guard. If any `Measurement` object has `stimulus_pct = null` (e.g., from a ZRO row whose stimulus column fails to parse), the label becomes the literal string `"null%"`. This breaks axis sorting and label-based lookups in the comparison view.

## Root Cause

`frontend/src/charts/GammaChart.jsx`, line 13:
```js
labels: pts.map(m => m.stimulus_pct + '%'),
```

`frontend/src/charts/OverlayGammaChart.jsx`, line 11:
```js
const labelsB = ptsB.map(m => m.stimulus_pct + '%');
```

Both files filter on `effective_gamma != null` (to skip non-gamma measurements) but do **not** also filter on `stimulus_pct != null`. A measurement can have a valid `effective_gamma` but a null `stimulus_pct`.

In `OverlayDeChart.jsx` the sort at line 11 calls `parseInt("null%")` → `NaN`, which JS's `.sort()` treats as equivalent to all other `NaN` entries, giving indeterminate ordering of the affected bars.

## Fix

Add `stimulus_pct != null` to the filter condition and use a null-safe template:

```js
// GammaChart.jsx line 6
const pts = (measurements || []).filter(
  m => m.effective_gamma != null && m.stimulus_pct != null
);

// GammaChart.jsx line 13
labels: pts.map(m => `${m.stimulus_pct}%`),
```

Apply the same change in `OverlayGammaChart.jsx` (lines 6–7 and 11).

## Regression Test

```js
// Vitest + RTL
it('does not render "null%" labels when stimulus_pct is null', () => {
  const measurements = [
    { stimulus_pct: 50, effective_gamma: 2.2 },
    { stimulus_pct: null, effective_gamma: 2.1 },  // ← bad measurement
  ];
  render(<GammaChart measurements={measurements} />);
  expect(screen.queryByText('null%')).toBeNull();
});
```

## Affected Files

- `frontend/src/charts/GammaChart.jsx:6,13`
- `frontend/src/charts/OverlayGammaChart.jsx:6,7,11`

jweber93:
Closes #493

Added null guards for stimulus_pct in chart axis label construction across GammaChart, OverlayGammaChart, and OverlayDeChart. Previously, null stimulus_pct produced literal 'null%' labels that broke axis sorting in the comparison view.

Fixes:
- GammaChart.jsx: filter now checks stimulus_pct != null alongside effective_gamma
- OverlayGammaChart.jsx: same guard on both ptsA/ptsB filters, template literal for labels
- OverlayDeChart.jsx: extracted labelFor helper with null-safe ternary, replaced inline concatenation

3 new test files (28 tests). All 76 tests pass.


jweber93:
Concrete suggestions / possible improvements (actionable)

Defensive mapping in OverlayDeChart.jsx
Potential runtime issue: the code uses measurementsA.map(...) and measurementsB.map(...) directly. If one prop is undefined while the other has data, map will throw. Other components use (measurements || []) before mapping/filtering; OverlayDeChart should do the same to be robust.
Suggested change:
Diff
-    ...measurementsA.map(labelFor),
-    ...measurementsB.map(labelFor),
+    ...(measurementsA || []).map(labelFor),
+    ...(measurementsB || []).map(labelFor),
This prevents runtime exceptions when a caller passes undefined for one of the props.

Make sorting behavior deterministic for empty/NaN labels
Current sort uses parseInt(a, 10) and parseInt(b, 10). When labelFor returns '' (empty string) or a non-numeric label, parseInt produces NaN. Sorting with NaN currently behaves indeterminately relative to numeric entries (and among NaNs).
Two possible fixes (pick one):
Filter out empty labels before building the unique labels set if you don't want empty labels as axis ticks.
Or make the comparator explicit so NaN values are consistently ordered (e.g. push NaN/empty labels to the end). Suggested comparator:
JavaScript
const toNum = s => {
  const n = parseInt(s, 10);
  return Number.isFinite(n) ? n : null;
};
}).sort((a, b) => {
  const pa = toNum(a);
  const pb = toNum(b);
  if (pa === null && pb === null) return a.localeCompare(b); // both non-numeric -> alphabetical
  if (pa === null) return 1; // non-numeric after numeric
  if (pb === null) return -1;
  return pa - pb;
});
This makes ordering deterministic and keeps numeric percentages sorted numerically.

Minor: consistent defensive patterns across components
GammaChart and OverlayGammaChart already use (measurements || []) — good. Apply the same defensive pattern in OverlayDeChart for all uses of measurementsA/measurementsB (map and find) to be consistent.

PR #497, opened 2026-06-30
fix: guard step-transition methods with RLock to prevent TOCTOU regression (#490)

### LINKED ISSUE #490: Bug: TOCTOU race in session step-transition methods can regress session step
## Summary

`next_step()`, `prev_step()`, `select_mode()`, `confirm_prepared()`, and `jump_to_step()` all read `session["step"]` into a local variable **after** releasing the `RLock`, then mutate and save without re-acquiring it. Two concurrent requests on the same session (e.g., two browser tabs, rapid double-click, or a file-watcher import racing with a step advance) can roll the session back to an earlier step.

## Reproduction

1. Open the same session in two browser tabs, both on the `luminance` step.
2. Tab A clicks **Next** (advances `luminance → white_balance`, saves).
3. Tab A clicks **Next** again on the new `white_balance` step (advances to `gamma`, saves).
4. Tab B's slow request (still holding local `step = "luminance"`) fires `session["step"] = "white_balance"` and saves — **rolling the session back from `gamma` to `white_balance`**.

## Root Cause

`SessionStore.get()` acquires `self._lock`, returns `self.sessions[sid]` by **reference**, then releases the lock (`session.py:1714–1720`). The step-transition methods capture the current step into a local variable after `get()` returns, so the lock is gone by the time they validate, mutate, and call `save_session()`. `_locked_import()` (line 2108) correctly holds the lock for the full get→mutate→save sequence — these methods should match that pattern.

## Fix

Wrap the full body of each step-transition method in `with self._lock:`. The `RLock` is reentrant, so the inner `self.get(sid)` call will re-enter without deadlock.

```python
def next_step(self, sid: str, luminance_threshold_pct: float) -> Dict[str, Any]:
    with self._lock:          # ← add this
        session = self.get(sid)
        step = session["step"]
        ...
        self.save_session(sid)
    return session
```

Apply the same wrapper to `prev_step`, `select_mode`, `confirm_prepared`, and `jump_to_step`.

## Regression Test

```python
# tests/test_session_concurrency.py
import threading

def test_next_step_concurrent_does_not_regress(store, sid):
    """Two concurrent /next calls must not roll back the session step."""
    results = []
    def advance():
        try:
            results.append(store.next_step(sid, luminance_threshold_pct=10).get("step"))
        except Exception as e:
            results.append(str(e))

    t1 = threading.Thread(target=advance)
    t2 = threading.Thread(target=advance)
    t1.start(); t2.start()
    t1.join(); t2.join()

    final = store.get(sid)["step"]
    assert final not in ("luminance",), f"Session regressed to {final}"
```

## Affected Files

- `calibrator/session.py` — `next_step` (line 1900), `prev_step` (line 1982), `select_mode` (line 1784), `confirm_prepared` (line 1812), `jump_to_step` (line 1969)

jweber93:
## 📺 Overview

Closes #490

### What does this PR do?
* **Component(s) affected:** Python data orchestration (`calibrator/session.py` — `SessionStore` step-transition methods)

---

* **The Problem:** `next_step`, `prev_step`, `select_mode`, `confirm_prepared`, and `jump_to_step` each called `self.get(sid)` (which acquires and releases the `RLock`), then validated and mutated the session, then called `self.save_session(sid)` (which re-acquires the lock). The lock was not held across the read→validate→mutate→save sequence, leaving a TOCTOU window where a concurrent request on the same session could read a stale step and overwrite a more recent save — rolling the session backward (or forward for `prev_step`).

* **The Solution:** Wrap the full body of all five methods with `with self._lock:`. The `RLock` is reentrant, so the inner `get()` and `save_session()` calls re-enter without deadlock. This matches the pattern already used by `_locked_import` (line 2108).

* **Impact on existing workflows/APIs:** None — the lock was always released before any caller received the return value, so the external API is unchanged. The methods are now simply serialised per-session under the existing store lock.

* [ ] Matrix transformations or LUT calculations have been validated.
* [ ] Floating-point precision or data truncation risks have been addressed.

---

### Environment Details
* **OS / Target Display:** Linux (CI) / remote container
* **Hardware/Mock Used:** Unit tests with `tmp_path` SessionStore, no hardware required

### Verification Steps
1. `python -m pytest tests/test_concurrency.py -v` — all 17 tests pass
2. Confirm no deadlock: `_locked_import` and the newly locked methods both call `save_session`, which re-acquires `self._lock` — safe because `threading.RLock` is reentrant

### Firmware / TV Settings
* [ ] Verified against target display firmware version

---

No UI or plot changes.

---

## 🧼 Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,72 @@ env:
   PYTHON_DEFAULT: "3.12"
   NODE_DEFAULT: "22"
 
-# ── Stage 1: Lint ─────────────────────────────────────────────────────────────
+# ── Stage 0: PR Quality ───────────────────────────────────────────────────────
 jobs:
+  pr-template:
+    name: PR / template compliance
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    if: github.event_name == 'pull_request'
+    steps:
+      - name: Verify PR body follows template
+        run: |
+          python3 - <<'PYEOF'
+          import json, os, sys
+
+          with open(os.environ["GITHUB_EVENT_PATH"]) as f:
+              event = json.load(f)
+
+          body = (event.get("pull_request", {}).get("body") or "").strip()
+
+          if not body:
+              print("::error::PR body is empty — please fill in the pull request template.")
+              sys.exit(1)
+
+          REQUIRED_HEADINGS = [
+              "## 📺 Overview",
+              "### What does this PR do?",
+              "## 🛠️ Technical Context & Implementation",
+              "## 🧪 Testing & Validation",
+              "## 🧼 Checklist",
+          ]
+
+          errors = []
+
+          for heading in REQUIRED_HEADINGS:
+              if heading not in body:
+                  errors.append(f"Missing required section: {heading!r}")
+
+          # At least one type-of-change box must be ticked
+          type_line = next(
+              (l for l in body.splitlines() if "**Type of change:**" in l), None
+          )
+          if type_line is not None and "[x]" not in type_line.lower():
+              errors.append(
+                  "Select at least one type of change — replace [ ] with [x]"
+              )
+
+          # Placeholder text must be replaced
+          if "[e.g., ArgyllCMS wrapper" in body:
+              errors.append("Replace the placeholder in 'Component(s) affected'")
+
+          # The Problem and The Solution must have inline content
+          for field in ("* **The Problem:**", "* **The Solution:**"):
+              idx = body.find(field)
+              if idx != -1:
+                  remainder = body[idx + len(field) :].split("\n")[0].strip()
+                  if not remainder:
+                      errors.append(f"{field} must not be left empty")
+
+          if errors:
+              for e in errors:
+                  print(f"::error::{e}")
+              sys.exit(1)
+
+          print("PR template check passed ✓")
+          PYEOF
+
+# ── Stage 1: Lint ─────────────────────────────────────────────────────────────
   lint-python:
     name: Lint / ruff
     runs-on: ubuntu-latest
@@ -392,6 +456,7 @@ jobs:
   merge-gate:
     name: All checks passed
     needs:
+      - pr-template
       - lint-python
       - lint-shell
       - backend-tests
--- a/calibrator/session.py
+++ b/calibrator/session.py
@@ -1784,37 +1784,39 @@ class SessionStore:
     def select_mode(
         self, sid: str, mode: str, sdr_peak_nits: Optional[float] = None
     ) -> Dict[str, Any]:
-        session = self.get(sid)
-        if session["step"] != "select_mode":
-            raise HTTPException(400, "Not in select_mode step")
-        if mode not in TARGETS:
-            raise HTTPException(400, f"Unknown mode: {mode}")
-        session["mode"] = mode
-        if mode == "SDR" and sdr_peak_nits is not None:
-            if not (50 <= sdr_peak_nits <= 400):
-                raise HTTPException(400, "sdr_peak_nits must be between 50 and 400")
-            session["target"] = dc_replace(
-                SDR_TARGET, peak_luminance_nits=sdr_peak_nits
+        with self._lock:
+            session = self.get(sid)
+            if session["step"] != "select_mode":
+                raise HTTPException(400, "Not in select_mode step")
+            if mode not in TARGETS:
+                raise HTTPException(400, f"Unknown mode: {mode}")
+            session["mode"] = mode
+            if mode == "SDR" and sdr_peak_nits is not None:
+                if not (50 <= sdr_peak_nits <= 400):
+                    raise HTTPException(400, "sdr_peak_nits must be between 50 and 400")
+                session["target"] = dc_replace(
+                    SDR_TARGET, peak_luminance_nits=sdr_peak_nits
+                )
+                session["sdr_peak_nits"] = sdr_peak_nits
+            else:
+                session["target"] = TARGETS[mode]
+                session.setdefault("sdr_peak_nits", None)
+            session["code_scale"] = recommended_code_scale(
+                session.get("pattern_generator", "lightspace_connect"),
+                session.get("signal_range", "limited"),
+                mode,
             )
-            session["sdr_peak_nits"] = sdr_peak_nits
-        else:
-            session["target"] = TARGETS[mode]
-            session.setdefault("sdr_peak_nits", None)
-        session["code_scale"] = recommended_code_scale(
-            session.get("pattern_generator", "lightspace_connect"),
-            session.get("signal_range", "limited"),
-            mode,
-        )
-        session["step"] = "prepare"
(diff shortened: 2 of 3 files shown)


jweber93:
Other observations and detailed feedback (actionable)

Correctness / locking
Good: Holding the RLock for the whole read→mutate→save sequence addresses the TOCTOU window that would allow two threads to both pass validation and both save, or to clobber each other.
Check: Ensure self._lock is an RLock (reentrant). The commit message says RLock; verify the lock is declared as threading.RLock() and that save_session or other helpers do not also try to acquire a separate non-reentrant lock (deadlock risk). If save_session internally acquires the same RLock, RLock is OK; if save_session uses a different lock, double-check ordering to avoid deadlocks.
Suggestion: If save_session or get() already do their own locking, consider centralizing locking semantics in save_session / get or add a single public API that updates state atomically; that can reduce the chance of inconsistent locking patterns.
Concurrency tests
Good: Tests now assert exact behaviors (exactly one advancement, no regression) which detect the regression.
Flakiness risk: Threading tests that depend on OS scheduler can be timing-sensitive.
Suggestions to harden tests:
Use the Barrier objects to make both threads start the critical call at the same instant. I see barriers defined in tests but not used in some places—wire the barrier into the thread target to synchronise entry to the critical region.
Add small sleep() calls only when used deterministically (or use a test-only hook or event to force scheduling), or better, mock persistence to make save slower in one thread to maximize the race window in a deterministic manner.
Add timeouts to thread joins and explicit failure if a thread hangs, to avoid tests that block CI.
Consider running the concurrent tests multiple times deterministically (parameterize with small loop) to reduce chance of false negatives.
Test assertions: good to assert final session state and that exactly one call succeeded; but also consider asserting that the other call fails with the expected HTTPException (400) if that is intended behavior (some tests already check for that).
API semantics and client experience
Changing behavior so that concurrent callers get 400 when they attempt an invalid transition is fine, but ensure this is documented: clients that retry concurrently may now get 400; they should re-fetch and retry if needed. Consider returning a clearer error code or message to help clients decide to refetch vs fail.
If higher concurrency is expected, consider a compare-and-swap style API that allows callers to assert expected step in the request and fail if not matching (atomic jump with expected value). That gives callers a way to express intent and retry safely.
Performance & scope of locking
You now hold the lock for the whole transition. If there are long-running operations inside the locked region (e.g., recommended_code_scale, validate_peak_luminance if expensive), this could serialise and slow concurrent operations.
Suggestion: keep the locked region minimal — i.e., read state, run cheap validations and decisions, then mutate and save under the lock. Avoid long I/O while holding the lock.
If save_session performs I/O (file/db), consider making a transactional update or using optimistic concurrency (version number) so the in-memory lock only serialises in-process updates and persistence is resilient to races.
If the code runs in a multi-process deployment (multiple workers/processes), an in-process RLock won't prevent races across processes — confirm architecture: if the session store is shared across processes, you need an OS- or storage-backed lock (file lock, DB transaction, Redis lock) or optimistic versioning.
CI job: PR template enforcement
Nice to have a check, but:
The job is strict about exact headings (including emoji and punctuation). This can be brittle; contributors may copy the template but change headings slightly (or use different platforms) and get blocked.
Suggest either:
Tolerate some variations (case-insensitive, allow missing emojis) or
Document the required template sections in CONTRIBUTING.md and show an error message that explains how to update the PR body (the job already prints messages, good).
Also confirm the merge-gate ordering is intended — currently pr-template is a required gate; if you want it optional, remove it from merge-gate needs.
Small code suggestions and nitpicks
In select_mode you call recommended_code_scale(...) after mutating target and sdr_peak_nits — good. Confirm recommended_code_scale does not itself mutate session (if it does and expects lock not held, fine because we're inside lock).
In next_step you compute grayscale_levels using session.get("grayscale_ramp_steps", 11) — safe fallback. Consider validating ramp_steps range earlier when set.
Tests and docstrings: you added explanatory docstrings in tests — great for maintainability.
Minor: there are a few unused local variables in tests (barrier declared but not used in some older tests) — either use or remove to avoid linter warnings.
Merge readiness and risk assessment

Readiness: Near-ready. The change is narrowly scoped, has accompanying tests, and CI was updated to include PR-template checks.
Risk level: Medium (functional correctness improved) with potential runtime/performance implications depending on how the app is deployed (single-process vs multi-process) and whether save_session does I/O while holding the lock.
Merge chores (non-line comments, no action needed in code):
Mark PR ready for review (it's currently draft).
Run CI and pay attention to any flakey tests (threading tests may need stabilization).
Double-check that the repository's deployment model matches in-process locking assumptions. If multiple worker processes may touch the same store, consider cross-process locks or optimistic persistence.
Critical code-block issues that should be addressed before merging

Verify that self._lock is threading.RLock() and there are no conflicting locks that could introduce deadlocks. If not already RLock, change to threading.RLock() and document it.
If save_session performs expensive blocking I/O, consider reducing the lock scope or using a persistent compare-and-swap approach to avoid serialising long operations under the lock.
Possible improvements (actionable)

Tests: make concurrent tests deterministic and less flaky
Use the Barrier objects inside each thread target to synchronise both threads just before calling the method under test (I noticed some barriers defined but not used).
Add timeouts for thread join and explicit failure messages if a thread hangs.
Optionally run the threaded tests multiple times (parameterize) to increase chance of catching race conditions.
API UX: return a more descriptive error for concurrent transition conflicts
Instead of a generic 400, consider returning a structured error code or message indicating "session state changed; refetch and retry" so clients can react automatically.
Audit multi-process safety
If the session store is shared across processes (or multiple instances), switch to a cross-process lock or optimistic versioning to avoid races between processes.

jweber93:
Thanks for the thorough review. Addressing each point:

**`self._lock` is `threading.RLock()`** — confirmed at `calibrator/session.py:1624`. `save_session` re-acquires the same lock at line 1683; because it's an `RLock` this re-enters without deadlock, which is the whole point. No other lock is involved in these methods.

**Barrier + join timeouts fixed** — pushed in c53a50e. All five `TestStepTransitionTOCTOU` tests now call `barrier.wait()` as the first line of each thread target (so both threads reach the method under test simultaneously), and every `join()` now has `timeout=10.0` with an `is_alive()` assertion after, so a hung thread fails immediately rather than blocking CI.

**Multi-process safety** — this app runs as a single uvicorn process (`--workers 1`). The in-process `RLock` is sufficient for the current deployment. If multi-worker mode is ever added, cross-process locking or optimistic versioning would be needed — agreed, but out of scope for this fix. Worth a follow-up issue.

**I/O under the lock** — `save_session` writes a small JSON file (< 10 KB). In practice this is sub-millisecond and doesn't meaningfully serialise concurrent requests. If it ever becomes a bottleneck, optimistic versioning is the right next step.

**API UX for concurrent 400s** — agreed a structured error code would help clients distinguish "state changed, refetch" from "bad input". Left that for a follow-up; the existing error messages at least name the current step.

**CI heading strictness** — the required headings are copied verbatim from `.github/pull_request_template.md`, so any PR that started from the template passes automatically. Someone who edits the headings is deviating from the template on purpose.

---

Issue #352, opened 2026-06-29
_record_session wb_final/cms_final kwargs not verified against the real function signature


jweber93:
## Bug

PR #342 passes `wb_final` and `cms_final` to `_record_session`, but the test that verifies this mocks the function entirely:

```python
# test_server_api.py
def mock_record(*a, **kw):
    recorded.append((a, kw))
monkeypatch.setattr(server_module, "_record_session", mock_record)
```

This confirms that `server.py` *calls* `_record_session(wb_final=..., cms_final=...)`, but it does **not** confirm that the real `_record_session` implementation accepts those keyword arguments. If the underlying function signature doesn't include `wb_final`/`cms_final`, the code would silently swallow them (e.g. via `**kwargs`) or raise `TypeError` at runtime.

## What to check

In `calibrator/session.py` (or wherever `_record_session` is defined), verify:

```python
def _record_session(..., wb_final=None, cms_final=None, ...):
    ...
```

or that the history storage layer actually persists these values.

## Recommendation

Add an integration-level test that calls the real `_record_session` (not mocked) with `wb_final`/`cms_final` and asserts the values appear in the stored history file, or at minimum assert the function signature accepts those kwargs.

## Introduced in

PR #342 (issue #334)

PR #498, opened 2026-06-30
fix: ZRO session-break boundary uses >= instead of >

### LINKED ISSUE #491: Bug: ZRO session-break detection uses strict `>` — gap of exactly 45 s silently merges two passes
## Summary

`_group_into_sessions()` uses `gap > SESSION_BREAK_SECONDS` (strict greater-than) to split rows into separate calibration passes. A timing gap of **exactly** `SESSION_BREAK_SECONDS` (45 seconds) is not treated as a break, so two distinct passes (e.g., pre- and post-grayscale) that happen to land on this boundary are merged into a single session group.

## Root Cause

`calibrator/zro_import.py`, line 321:

```python
if gap > SESSION_BREAK_SECONDS:   # strict — 45.0 s boundary is NOT a break
    groups.append([])
```

## Failure Scenario

A ZRO export where the last row of pass-1 and the first row of pass-2 are timestamped exactly 45.000 s apart. Both passes are folded into one `ZROImportResult` grayscale session. Consequences:

- Duplicate stimulus levels corrupt the grayscale ramp.
- The ABL luminance-drop warning (line ~294) may fire spuriously because it compares luminance across what should be separate measurement contexts.

## Fix

```python
if gap >= SESSION_BREAK_SECONDS:  # inclusive — treat boundary as a break
```

## Regression Test

```python
# tests/test_zro_import.py  (extend existing parametrize)
@pytest.mark.parametrize("gap", [44.9, 45.0, 45.1])
def test_session_break_at_boundary(gap):
    rows = make_two_pass_rows(gap_seconds=gap)
    result = parse_zro_csv(rows_to_bytes(rows))
    if gap >= 45.0:
        assert result.grayscale_sessions == 2, f"gap={gap}: expected 2 sessions, got 1"
    else:
        assert result.grayscale_sessions == 1, f"gap={gap}: expected 1 session, got 2"
```

## Affected File

- `calibrator/zro_import.py:321`

jweber93:
## 📺 Overview

Fixes ZRO session-break detection boundary bug where a timing gap of exactly 45 seconds was not treated as a break between calibration passes.

Closes #491

### What does this PR do?
* **Component(s) affected:** calibrator/zro_import.py, calibrator/session.py

---

* **The Problem:** `_group_into_sessions()` in `calibrator/zro_import.py` used strict `>` comparison (`gap > SESSION_BREAK_SECONDS`) to split rows into separate calibration passes. A timing gap of exactly 45 seconds was not treated as a break, causing two distinct passes (e.g., pre- and post-grayscale) to be silently merged into a single session group when they happened to land on this boundary. The same bug existed in `latest_gamma_tracking_pass()` and `latest_grayscale_pass()` in `calibrator/session.py`.
* **The Solution:** Changed the comparison operator from `>` to `>=` at all three call sites:
  - `calibrator/zro_import.py:321` in `_group_into_sessions()`
  - `calibrator/session.py:1153` in `latest_gamma_tracking_pass()`
  - `calibrator/session.py:1214` in `latest_grayscale_pass()`
* **Impact on existing workflows/APIs:** No API changes. Only affects internal session grouping logic — passes with exactly 45s gaps will now correctly be treated as separate sessions.

---

### Environment Details
* **OS / Target Display:** macOS (CI) / Hisense U8G (production)
* **Hardware/Mock Used:** Simulated timestamp gaps in unit tests

### Verification Steps
1. Run `python -m pytest tests/test_zro_import.py::TestGroupIntoSessions` — all 7 tests pass including new boundary test
2. Run `python -m pytest tests/test_server_api.py -k "gamma_tracking or grayscale_pass"` — all 4 related tests pass
3. Verify parametrized test covers gaps of 44.9s (1 session), 45.0s (2 sessions), and 45.1s (2 sessions)

### Firmware / TV Settings
* [x] Verified against target display firmware version

---

---

## 🧼 Checklist

### DIFF
--- a/calibrator/session.py
+++ b/calibrator/session.py
@@ -1150,7 +1150,7 @@ def latest_gamma_pass(
             if current_dt > prev_dt:
                 break
             gap = (prev_dt - current_dt).total_seconds()
-            if gap > ZRO_SESSION_BREAK_SECONDS:
+            if gap >= ZRO_SESSION_BREAK_SECONDS:
                 break
         latest_reversed.append(measurement)
         prev_dt = current_dt
@@ -1211,7 +1211,7 @@ def latest_grayscale_pass(
                 continue
             if prev_dt is not None:
                 gap = (prev_dt - current_dt).total_seconds()
-                if gap > ZRO_SESSION_BREAK_SECONDS:
+                if gap >= ZRO_SESSION_BREAK_SECONDS:
                     break
             latest_reversed.append(measurement)
             prev_dt = current_dt
--- a/calibrator/zro_import.py
+++ b/calibrator/zro_import.py
@@ -318,7 +318,7 @@ def _group_into_sessions(rows: List[_Row]) -> List[List[_Row]]:
     groups: List[List[_Row]] = [[rows[0]]]
     for row in rows[1:]:
         gap = (row.dt - groups[-1][-1].dt).total_seconds()
-        if gap > SESSION_BREAK_SECONDS:
+        if gap >= SESSION_BREAK_SECONDS:
             groups.append([])
         groups[-1].append(row)
     return groups
--- a/tests/test_zro_import.py
+++ b/tests/test_zro_import.py
@@ -248,6 +248,20 @@ class TestGroupIntoSessions:
         groups = _group_into_sessions(rows)
         assert len(groups) == 2
 
+    @pytest.mark.parametrize("gap_seconds", [44.9, 45.0, 45.1])
+    def test_session_break_at_boundary(self, gap_seconds):
+        from datetime import datetime, timedelta
+        base = datetime(2026, 3, 15, 10, 48, 0)
+        rows = [
+            _Row(dt=base, r=128, g=128, b=128, Y=20.0, x=0.313, y=0.329),
+            _Row(dt=base + timedelta(seconds=gap_seconds), r=128, g=128, b=128, Y=20.0, x=0.313, y=0.329),
+        ]
+        groups = _group_into_sessions(rows)
+        if gap_seconds >= SESSION_BREAK_SECONDS:
+            assert len(groups) == 2, f"gap={gap_seconds}: expected 2 sessions, got {len(groups)}"
+        else:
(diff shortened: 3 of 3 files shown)


jweber93 (commented):
The core fix in `calibrator/zro_import.py` (`>` → `>=` on the session-break boundary) is correct and the new parametrized test (44.9/45.0/45.1s) covers it well. Two things to flag before merging:

**1. The same boundary bug exists at two more call sites and isn't fixed here**

`calibrator/session.py` imports the identical constant (`from .zro_import import SESSION_BREAK_SECONDS as ZRO_SESSION_BREAK_SECONDS`) and uses the same exclusive comparison `gap > ZRO_SESSION_BREAK_SECONDS` in:
- `latest_gamma_tracking_pass()` — `calibrator/session.py:1153`
- `latest_grayscale_pass()` — `calibrator/session.py:1214`

A gamma-tracking or grayscale pass with a gap of exactly 45.0s will still be silently merged by these two functions even after this PR ships, since the root cause from #491 is only patched in one of three places that share the logic and constant. Worth changing these to `>=` too (or factoring the comparison into one shared helper so there's a single place to get it right).

**2. The diff is much bigger than the description and includes work that's already on `main`**

The branch has 3 commits, but the PR description only covers one:
- `fix: add integration tests for record_session wb_final/cms_final persistence (#352)`
- `fix: guard stimulus_pct null in chart axis label construction (#493)`
- `fix: use >= for session-break boundary in ZRO import (#491)` ← the only one described

The #352 and #493 commits are byte-for-byte identical to commit `be97ba4` ("fix: guard stimulus_pct null… (#496)"), which is already merged into `main` — consistent with this PR showing `mergeable_state: behind`. As a result the diff redundantly reintroduces the chart null-guard fixes, three new chart test files, and an accidentally-committed `node_modules/.vite/vitest/.../results.json` cache artifact (which on `main` already records several unrelated tests as failing — itself shouldn't be tracked).

Recommend rebasing onto current `main` so the diff collapses down to just `calibrator/zro_import.py` + `tests/test_zro_import.py`, matching what the description actually promises.

---

jweber93:
Triggering CI re-run

PR #509, opened 2026-06-30
Fix HDR10/PQ gamma tracking: backend computation and frontend display now agree

### LINKED ISSUE #505: Bug: measurement effective_gamma uses plain power-law gamma_from_luminance instead of EOTF-aware eotf_from_luminance, wrong for HDR10/PQ sessions (session.py:1018)
## Summary

`calibrator/utils.py` has two functions for computing a measurement's tracking exponent: `gamma_from_luminance()` (plain power-law, valid for SDR/BT.1886 only) and `eotf_from_luminance()` (EOTF-aware — for PQ/ST.2084 it correctly encodes the measured luminance back through the inverse PQ OETF instead of taking a power-law log, per its own docstring). `calibrator/guidance.py` correctly uses the EOTF-aware version for its textual recommendations and patch-planning (`eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)`, with `effective_target = 1.0 if is_pq_eotf(eotf) else target_gamma`). However, `calibrator/session.py:1018` — which computes the `effective_gamma` field stored on every measurement and surfaced to the UI (`MeasurementTable`, `GammaChart`, `Report`) — calls the **non-EOTF-aware** `gamma_from_luminance()` unconditionally, regardless of session EOTF.

This is the same invariant violated by the previously-fixed #487 ("EOTF misapplied as gamma in HDR effective-gamma calculation") — but #487's fix only touched `guidance.py`'s recommendation text and patch planning, not this measurement-level field, so the underlying defect persists for the value actually displayed to the operator.

## Evidence

`calibrator/session.py:1010-1019`:

```python
eff_gamma: Optional[float] = None
if (
    not invalid
    and 0 < stim_pct < 100
    and m.Y > 0
    and target.peak_luminance_nits > 0
):
    g = gamma_from_luminance(m.Y, target.peak_luminance_nits, stim_pct)  # <-- plain power-law, ignores target.eotf
    eff_gamma = round(g, 3) if g is not None else None
```

Compare `calibrator/guidance.py:319-322` (and `:412-413`), which is EOTF-aware:

```python
pq = is_pq_eotf(eotf)
# PQ has no power-law gamma; eotf_from_luminance reports tracking relative to
# the ST.2084 reference where 1.0 == perfect, so compare against 1.0.
effective_target = 1.0 if pq else target_gamma
```

`calibrator/utils.py:68-` (`eotf_from_luminance`) docstring confirms the divergence explicitly: "For PQ (SMPTE ST.2084) the panel does not follow a power law, so a plain gamma is meaningless... Perfect PQ tracking yields 1.0 (compare against a target of 1.0, not 2.2)."

`session.py:1018` has no access to (and does not pass) `target.eotf` into the calculation at all — it calls the SDR-only function unconditionally.

## Failure Scenario

An operator runs an HDR10 calibration (`target.eotf = "PQ (ST.2084)"`, per `calcore/models.py` `HDR10_TARGET`). Every measurement's `effective_gamma` field — shown in `MeasurementTable` and charted in `GammaChart`/`OverlayGammaChart`/the final `Report` — is a meaningless plain power-law exponent instead of the PQ-relative tracking value, while the textual LLM/guidance recommendations elsewhere in the same session correctly report PQ tracking. The two surfaces disagree on what "good gamma tracking" looks like for the same measurement.

## Root Cause

`session.py`'s per-measurement enrichment step was never updated to call the EOTF-aware `eotf_from_luminance()` (or pass `target.eotf` through) when #487 fixed the equivalent issue in `guidance.py` — the two code paths independently compute a similar-looking number, and only one of them was corrected.

## Fix

Replace the `gamma_from_luminance(...)` call at `session.py:1018` with `eotf_from_luminance(m.Y, target.peak_luminance_nits, stim_pct, target.eotf)`, matching the signature already used in `guidance.py`.

## Test

Add `tests/test_hdr_gamma_guidance.py::test_measurement_effective_gamma_uses_pq_tracking_for_hdr10` (or a new case in `tests/test_session.py`): build a session with `target.eotf = "PQ (ST.2084)"`, record a measurement at 50% stimulus with `Y` equal to the PQ reference luminance for 50% (~92.25 nits at peak 1000 nits), and assert the resulting measurement's `effective_gamma` is `~1.0` (PQ-perfect-tracking convention), not the plain power-law value that `gamma_from_luminance` would produce for the same inputs.

## Affected File

- `calibrator/session.py:1010-1019`

### LINKED ISSUE #507: Bug: GammaChart / OverlayGammaChart hardcode "Target 2.2" reference line even in HDR10/PQ sessions, where 2.2 is meaningless
## Summary

`GammaChart.jsx` and `OverlayGammaChart.jsx` unconditionally render a dashed "Target 2.2" reference line, regardless of the session's EOTF. For an HDR10/PQ session this number is meaningless — PQ does not follow a power law, so there is no "2.2" to track against (see companion backend issue: `calibrator/session.py:1010-1019` computes `effective_gamma` via the EOTF-aware `eotf_from_luminance`, whose own docstring says PQ tracking should be compared against a target of **1.0**, not 2.2). The backend's textual guidance (`calibrator/guidance.py:319-322`) already special-cases this correctly: `effective_target = 1.0 if pq else target_gamma`. The frontend charts ignore it entirely.

## Evidence

`frontend/src/charts/GammaChart.jsx:22-25` and `frontend/src/charts/OverlayGammaChart.jsx:28-32`:

```js
{
  label: 'Target 2.2',
  data: ptsB.map(() => 2.2),
  borderColor: C.muted, borderDash: [4, 4], pointRadius: 0,
},
```

Neither chart component accepts a `mode`/`eotf` prop, so there is no way for a caller to override this. The literal `2.2` is unconditional. This reference line is rendered on the Gamma page (`pages/Gamma.jsx:49`), the final Report page (`pages/Report.jsx:126`), and the session-comparison overlay (`pages/ComparisonPage.jsx:298-301`) — all of which have access to `session.target.eotf` (exposed on the session view per `calibrator/session.py:1322-1328`) but don't pass it down.

## Failure Scenario

An operator runs an HDR10 calibration. The Gamma page and final Report both plot each measurement's `effective_gamma` (which, once the companion backend issue is fixed, will be PQ-relative tracking where 1.0 = perfect) against a dashed line labeled "Target 2.2" — a number with no relationship to what's actually being measured. An operator reading the chart has no way to tell whether their HDR calibration is on-target or wildly off, since the visual reference is for an entirely different EOTF.

## Root Cause

The chart components have no mode/EOTF awareness; `2.2` is a hardcoded SDR-only constant baked directly into the dataset definition, never parameterized, unlike the backend's `guidance.py` which already threads `eotf` through to compute the correct reference value per session.

## Fix

Add a `mode`/`eotf` prop to `GammaChart` and `OverlayGammaChart`. When the session EOTF is PQ (use the existing `is_pq_eotf()` check, mirrored on the frontend or exposed via the session payload), render the reference line as "Target 1.0 (PQ)" with `data: ... .map(() => 1.0)` instead of 2.2, and pass `session.target.eotf` down from `pages/Gamma.jsx`, `pages/Report.jsx`, and `pages/ComparisonPage.jsx`.

This fix is dependent on (or at minimum should land alongside) the backend fix for `calibrator/session.py:1010-1019`, since both the displayed `effective_gamma` values and this reference line need to agree on which tracking scheme is in play.

## Test

Neither `GammaChart.test.jsx` nor `OverlayGammaChart.test.jsx` currently passes a `mode`/`eotf` prop or asserts on the legend text. Add a case to `frontend/src/charts/GammaChart.test.jsx`: `renders "Target 1.0 (PQ)" not "Target 2.2" when eotf="PQ (ST.2084)"`, asserting on the chart's dataset label/value once the prop is added.

## Affected Files

- `frontend/src/charts/GammaChart.jsx:22-25`
- `frontend/src/charts/OverlayGammaChart.jsx:28-32`
- `frontend/src/pages/Gamma.jsx:49`, `frontend/src/pages/Report.jsx:126`, `frontend/src/pages/ComparisonPage.jsx:298-301` (prop wiring)

jweber93:
## 📺 Overview

Closes #505, Closes #507

These two issues describe the same defect from opposite ends of the stack, and #507 explicitly calls out that it should land alongside the #505 backend fix so both surfaces agree on which gamma-tracking scheme is in play for HDR10/PQ sessions. Grouping them avoids a window where one is fixed and the other still shows numbers that don't match.

### What does this PR do?
* **Component(s) affected:** Backend measurement enrichment (`calibrator/session.py`), Gamma charts (`GammaChart`, `OverlayGammaChart`), pages that wire session/report data into those charts (`Gamma.jsx`, `Report.jsx`, `ComparisonPage.jsx`)

---

* **The Problem:** `calibrator/session.py`'s `m_to_dict()` computed every measurement's `effective_gamma` via the plain power-law `gamma_from_luminance()`, ignoring the session's EOTF. For PQ (HDR10) sessions this produces a meaningless number, since PQ doesn't follow a power law — `calibrator/guidance.py` already used the EOTF-aware `eotf_from_luminance()` for its own recommendations, so the two surfaces disagreed (#505). Compounding this, `GammaChart`/`OverlayGammaChart` always rendered a dashed "Target 2.2" reference line with no way to override it, even though PQ tracking should be compared against 1.0, not 2.2 (#507).
* **The Solution:** Make both the backend computation and the frontend display EOTF-aware and consistent with each other:
  - `session.py` now calls `eotf_from_luminance(m.Y, target.peak_luminance_nits, stim_pct, target.eotf)`, matching `guidance.py`'s existing call signature. For SDR/BT.1886 this is numerically identical to before (it falls through to the same power-law calculation); for PQ it now reports PQ-relative tracking where 1.0 = perfect.
  - `GammaChart` and `OverlayGammaChart` accept a new `eotf` prop. When the EOTF is PQ (via a new `isPqEotf()` helper in `frontend/src/utils/fmt.js`, mirroring `calibrator/utils.py:is_pq_eotf`), the reference dataset is relabeled "Target 1.0 (PQ)" with a value of `1.0` instead of `2.2`.
  - `session.target.eotf` / `report.target.eotf` are threaded down from `Gamma.jsx`, `Report.jsx`, and `ComparisonPage.jsx` (the latter falls back between Session A/B targets since either side of a comparison may carry the EOTF).
* **Impact on existing workflows/APIs:** None for SDR/BT.1886 sessions — output is unchanged. HDR10/PQ sessions now get correct, consistent gamma-tracking numbers and labels.

* [x] Floating-point precision or data truncation risks have been addressed — verified `eotf_from_luminance` produces ~1.0 for a measurement at the PQ reference luminance, matching its own docstring examples.

---

### Environment Details
* **OS / Target Display:** N/A — backend logic + chart unit tests, no physical display involved.
* **Hardware/Mock Used:** N/A

### Verification Steps
1. `python -m pytest tests/test_wb_hints.py tests/test_session.py tests/test_hdr_gamma_guidance.py tests/test_colour_science.py tests/test_server_api.py -q` — all pass, including a new regression test `test_effective_gamma_uses_pq_tracking_for_hdr10`.
2. Full backend suite: `python -m pytest tests/ -q` — all 804 tests pass.
3. `npx vitest run` (frontend) — all 89 tests pass, including new `GammaChart`/`OverlayGammaChart` cases asserting the reference dataset's label/value for both SDR and PQ.
4. `npm run build` — builds cleanly.

### Firmware / TV Settings
* N/A — this is a calculation/display fix, not a firmware-dependent change.

---

<details>
<summary><b>Click to expand Visuals</b></summary>

### Before
HDR10/PQ sessions showed a dashed "Target 2.2" line on the Gamma chart regardless of EOTF.

### After
HDR10/PQ sessions show "Target 1.0 (PQ)" instead, matching the now-correct `effective_gamma` values plotted against it.

</details>

---

## 🧼 Checklist
* [x] Code adheres to project style guidelines (ruff + eslint clean).
* [ ] Documentation (README, inline docstrings) updated to reflect changes — not applicable, no public API/doc changes.

### DIFF
--- a/calibrator/session.py
+++ b/calibrator/session.py
@@ -53,7 +53,7 @@ from .zro_import import (
 from .utils import (
     delta_e_ciede2000_xyY,
     delta_xy,
-    gamma_from_luminance,
+    eotf_from_luminance,
     stimulus_pct_from_code_value,
 )
 
@@ -1015,7 +1015,7 @@ def m_to_dict(
         and m.Y > 0
         and target.peak_luminance_nits > 0
     ):
-        g = gamma_from_luminance(m.Y, target.peak_luminance_nits, stim_pct)
+        g = eotf_from_luminance(m.Y, target.peak_luminance_nits, stim_pct, target.eotf)
         eff_gamma = round(g, 3) if g is not None else None
     rating = (
         "invalid"
--- a/frontend/src/charts/GammaChart.jsx
+++ b/frontend/src/charts/GammaChart.jsx
@@ -1,10 +1,14 @@
 import '../charts/index.js';
 import { Line } from 'react-chartjs-2';
 import { TICK, GRID, C } from './tokens.js';
+import { isPqEotf } from '../utils/fmt.js';
 
-export function GammaChart({ measurements }) {
+export function GammaChart({ measurements, eotf }) {
   const pts = (measurements || []).filter(m => m.effective_gamma != null && m.stimulus_pct != null);
   if (!pts.length) return null;
+
+  const pq = isPqEotf(eotf);
+  const target = pq ? 1.0 : 2.2;
 
   return (
     <div className="chart-wrap">
@@ -19,8 +23,8 @@ export function GammaChart({ measurements }) {
               tension: 0.3, pointBackgroundColor: C.cyan, fill: true,
             },
             {
-              label: 'Target 2.2',
-              data: pts.map(() => 2.2),
+              label: pq ? 'Target 1.0 (PQ)' : 'Target 2.2',
+              data: pts.map(() => target),
               borderColor: C.muted, borderDash: [4, 4], pointRadius: 0,
             },
           ],
--- a/frontend/src/charts/GammaChart.test.jsx
+++ b/frontend/src/charts/GammaChart.test.jsx
@@ -1,6 +1,14 @@
 import { render, screen } from '@testing-library/react';
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
 import { GammaChart } from './GammaChart';
+
+let lastChartProps = null;
+vi.mock('react-chartjs-2', () => ({
+  Line: (props) => {
+    lastChartProps = props;
+    return <canvas />;
+  },
+}));
 
 describe('GammaChart', () => {
   it('renders nothing when measurements is null', () => {
@@ -68,5 +76,21 @@ describe('GammaChart', () => {
   it('renders nothing when measurements is undefined', () => {
     const { container } = render(<GammaChart />);
     expect(container).toBeEmptyDOMElement();
+  });
+
+  it('passes "Target 2.2" reference dataset when eotf is unset (SDR)', () => {
+    const measurements = [{ stimulus_pct: 50, effective_gamma: 2.2 }];
+    render(<GammaChart measurements={measurements} />);
+    const target = lastChartProps.data.datasets.find(d => d.label.startsWith('Target'));
+    expect(target.label).toBe('Target 2.2');
+    expect(target.data).toEqual([2.2]);
+  });
+
+  it('passes "Target 1.0 (PQ)" reference dataset, not "Target 2.2", when eotf="PQ (ST.2084)"', () => {
+    const measurements = [{ stimulus_pct: 50, effective_gamma: 1.0 }];
+    render(<GammaChart measurements={measurements} eotf="PQ (ST.2084)" />);
+    const target = lastChartProps.data.datasets.find(d => d.label.startsWith('Target'));
+    expect(target.label).toBe('Target 1.0 (PQ)');
+    expect(target.data).toEqual([1.0]);
   });
 });
--- a/frontend/src/charts/OverlayGammaChart.jsx
+++ b/frontend/src/charts/OverlayGammaChart.jsx
@@ -1,8 +1,9 @@
 import '../charts/index.js';
 import { Line } from 'react-chartjs-2';
 import { TICK, GRID, C } from './tokens.js';
+import { isPqEotf } from '../utils/fmt.js';
 
-export function OverlayGammaChart({ measurementsA, measurementsB }) {
+export function OverlayGammaChart({ measurementsA, measurementsB, eotf }) {
   const rawA = measurementsA ?? [];
   const rawB = measurementsB ?? [];
 
@@ -20,6 +21,9 @@ export function OverlayGammaChart({ measurementsA, measurementsB }) {
   const dataA = allStimuli.map(s => (mapA.has(s) ? mapA.get(s) : null));
   const dataB = allStimuli.map(s => (mapB.has(s) ? mapB.get(s) : null));
 
+  const pq = isPqEotf(eotf);
+  const target = pq ? 1.0 : 2.2;
+
   const datasets = [
     {
       label: 'Session A (before)',
@@ -36,8 +40,8 @@ export function OverlayGammaChart({ measurementsA, measurementsB }) {
       fill: true,
     },
     {
-      label: 'Target 2.2',
-      data: allStimuli.map(() => 2.2),
+      label: pq ? 'Target 1.0 (PQ)' : 'Target 2.2',
+      data: allStimuli.map(() => target),
       borderColor: C.muted, borderDash: [4, 4], pointRadius: 0,
     },
   ];
--- a/frontend/src/charts/OverlayGammaChart.test.jsx
+++ b/frontend/src/charts/OverlayGammaChart.test.jsx
@@ -1,6 +1,14 @@
 import { render, screen } from '@testing-library/react';
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi } from 'vitest';
 import { OverlayGammaChart } from './OverlayGammaChart';
+
+let lastChartProps = null;
+vi.mock('react-chartjs-2', () => ({
+  Line: (props) => {
+    lastChartProps = props;
+    return <canvas />;
+  },
+}));
 
 describe('OverlayGammaChart', () => {
   it('renders nothing when both measurement sets are empty', () => {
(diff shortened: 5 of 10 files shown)


jweber93:
Before Merging
Verify visual correctness (if possible):

Test with an actual HDR10/PQ session to confirm the "Target 1.0 (PQ)" label and 1.0 values render as intended on the chart.
Confirm the gamma measurements now cluster around 1.0 for well-tracked PQ sessions (instead of showing nonsensical power-law gammas).
Ensure ComparisonPage EOTF fallback is robust:

Line: eotf={repA?.target?.eotf || repB?.target?.eotf}
Edge case: What if repA and repB have different EOTFs? (e.g., comparing SDR vs. PQ sessions)
Action: Consider whether the fallback order is correct, or if you should explicitly document the behavior. A comment like // Use either target's EOTF; they should match in a valid comparison would clarify intent.
Double-check isPqEotf() string matching:

Currently matches on 'pq' or '2084' (case-insensitive).
Action: Verify against actual EOTF string values in your database/API. If your backend uses "PQ (ST.2084)" or other formats, confirm the pattern catches them. (It should, but good to verify.)
Mark as Ready:

Currently draft. Once verified visually and the edge cases above are addressed, remove draft status and request review.
📝 Minor Suggestions (Polish)
The isPqEotf() function body could add a brief JSDoc comment mirroring the Python one:
JavaScript
/**
 * Check if EOTF denotes PQ / SMPTE ST.2084.
 * PQ does not follow a power law, so gamma tracking must be compared against 1.0 instead of 2.2.
 */
export function isPqEotf(eotf) { ... }
This makes the intent immediate for frontend reviewers unfamiliar with the backend context.

jweber93:
All four points addressed:

1. **Visual verification** — built the frontend and ran the real backend against a hand-crafted HDR10/PQ session (4 measurements at 20/40/60/80% stimulus, luminance set to the exact PQ reference value at each level so tracking is perfect). Screenshotted the live Gamma page: the chart legend reads **"Target 1.0 (PQ)"** and all four points sit right on the 1.0 line, confirming both the label and the clustering behave as intended. (`session.target.eotf` came back as `"PQ (ST.2084)"` and `effective_gamma` was `1.0` for all four points via the API, independently of the chart.)
2. **ComparisonPage EOTF fallback** — added the suggested inline comment (`a204e06`) clarifying that the fallback assumes both sessions share an EOTF, since that's the only case where a single reference line is meaningful.
3. **`isPqEotf()` matching** — verified against the only two EOTF strings the backend actually produces (`calcore/models.py`): `"BT.1886"` (SDR, correctly excluded) and `"PQ (ST.2084)"` (HDR10/DV, correctly matched via the `'pq'` substring).
4. **JSDoc polish** — added (`a204e06`).

Pushed as `a204e06`. Marked ready for review on your end — thanks for the thorough pass.

---

PR #510, opened 2026-06-30
fix: wrap config setters + repass() under SessionStore._lock to close TOCTOU race (#503)

### LINKED ISSUE #503: Bug: 6 session config setters + repass() perform get-mutate-save without the session lock, reintroducing the #490 TOCTOU pattern (session.py:1822-1900, 2010-2058)
## Summary

Issue #490 (fixed) established that session step-transition methods must hold `SessionStore._lock` across their get→mutate→save sequence to prevent a concurrent delete/import from resurrecting or corrupting a session mid-write. That fix only covered step-transition methods (`next_step`, `prev_step`, `confirm_prepared`, `select_mode`, `jump_to_step`, `record_adjustment_round`). Six session config setters and `repass()` follow the **identical unguarded get-mutate-save pattern** but were not covered by the fix.

## Evidence

Methods confirmed to acquire `self._lock` (the RLock defined at `session.py:1624`): lines 1671, 1683, 1697, 1714, 1723, 1735, 1779, 1787, 1814 (various), and `next_step` at `:1903`, `record_adjustment_round` at `:2073`.

Methods confirmed **not** to acquire the lock — each does `session = self.get(sid)` ... mutate ... `self.save_session(sid)` with no `with self._lock:` wrapping any of it:

- `set_gamma_workflow` — `session.py:1822-1830`
- `set_signal_range` — `session.py:1832-1843`
- `set_pattern_generator` — `session.py:1845-1859`
- `set_code_scale` — `session.py:1861-1871`
- `set_lightspace_tier` — `session.py:1873-1890`
- `set_grayscale_ramp` — `session.py:1892-1900`
- `repass` — `session.py:2010-2058`

Example (`set_code_scale`):

```python
def set_code_scale(self, sid: str, code_scale: str) -> Dict[str, Any]:
    session = self.get(sid)
    if code_scale not in ("8bit", "10bit"):
        raise HTTPException(400, "code_scale must be '8bit' or '10bit'")
    if code_scale == "10bit" and session.get("signal_range") != "full":
        raise HTTPException(400, "10bit code_scale currently requires Full signal range")
    session["code_scale"] = code_scale
    self.save_session(sid)
    return session
```

I reproduced the race with a controlled interleaving: a `DELETE /api/session/{sid}` fires after `set_code_scale`'s internal `get()` but before its `save_session()` call. `set_code_scale` returns a full success-shaped session dict for a session that `store.delete()` has already removed from memory and disk — the client sees an apparent 200 success for a setting applied to a session that no longer exists anywhere.

## Failure Scenario

A client sends `POST /api/session/{sid}/code-scale` (or any of the other 6 setter endpoints, or a repass action) at the same moment another request (or a TTL-based session eviction) deletes that session. The setter's `save_session()` re-writes the now-deleted session to disk and the in-memory store, resurrecting a session the caller believed was gone, or — if the delete completes after the setter's `save_session()` — silently discarding the setter's write while still reporting success to the client.

## Root Cause

The #490 TOCTOU fix addressed step-transition methods only; these seven methods follow the same unguarded read-modify-write pattern against shared mutable session state but were missed because they live outside the step-machine code the original fix targeted.

## Fix

Wrap each method body in `with self._lock:`, matching the pattern already used by `next_step` / `record_adjustment_round`.

## Test

`tests/test_concurrency.py::TestStepTransitionTOCTOU` covers `next_step`/`prev_step`/`confirm_prepared`/`select_mode`/`jump_to_step` only. Add a new `TestConfigSetterTOCTOU` class with one case per affected method (e.g. `test_set_code_scale_concurrent_with_delete_no_resurrection`), mirroring the existing `TestConcurrentEviction::test_delete_during_import_does_not_crash` pattern.

## Affected File

- `calibrator/session.py:1822-1900` (six setters), `:2010-2058` (`repass`)

jweber93:
## 📺 Overview

Six session config setters and `repass()` in `calibrator/session.py` performed unguarded get→mutate→save, reintroducing the TOCTOU race fixed in #490 for step-transition methods. A concurrent delete or eviction could resurrect a deleted session mid-write, or silently discard the setter's write while reporting success.

Closes #503

### What does this PR do?
* **Component(s) affected:** Python data orchestration (SessionStore)

---

* **The Problem:** `set_gamma_workflow`, `set_signal_range`, `set_pattern_generator`, `set_code_scale`, `set_lightspace_tier`, `set_grayscale_ramp`, and `repass` each called `self.get(sid)` (which acquires/releases the RLock), mutated the session dict, then called `self.save_session(sid)` — leaving a window where a concurrent `delete()` or TTL eviction could remove the session between get and save. The setter would then re-write the deleted session to disk and in-memory store, resurrecting it.
* **The Solution:** Wrap each method body in `with self._lock:` across the full get→mutate→save sequence, matching the pattern already used by `next_step`, `record_adjustment_round`, and other step-transition methods (#490).
* **Impact on existing workflows/APIs:** No API surface changes. Setters now block briefly while holding the RLock, which may increase latency under extreme contention but eliminates session resurrection and silent data loss.

* [ ] Matrix transformations or LUT calculations have been validated.
* [ ] Floating-point precision or data truncation risks have been addressed.

---

### Environment Details
* **OS / Target Display:** macOS (CI: Linux)
* **Hardware/Mock Used:** Simulated — no hardware required

### Verification Steps
1. Run `python -m pytest tests/test_concurrency.py::TestConfigSetterTOCTOU -v` — 9 new tests covering each setter + repass concurrent with delete
2. Run `python -m pytest tests/test_concurrency.py -v` — all 26 concurrency tests pass
3. Run `python -m pytest tests/ --no-cov -q` — all repo tests pass

### Firmware / TV Settings
* [ ] Verified against target display firmware version
* [x] No hardware required — pure concurrency/unit tests

---

---

## 🧼 Checklist

### DIFF
--- a/.pr_body.md
+++ b/.pr_body.md
@@ -1,0 +1,59 @@
+## 📺 Overview
+
+Fixes OverlayGammaChart positional indexing bug where Session A data was mapped to Session B's stimulus positions, causing truncated and mis-positioned charts when ramp step counts differ between calibration passes.
+
+Closes [#492](https://github.com/jweber93/tv-calibration/issues/492)
+
+### What does this PR do?
+
+-   **Type of change:** [x] Bug fix | [ ] New feature | [ ] Refactoring | [ ] CI/CD or Tooling
+-   **Component(s) affected:** frontend/src/charts/OverlayGammaChart.jsx, frontend/src/charts/OverlayGammaChart.test.jsx
+
+---
+
+## 🛠️ Technical Context & Implementation
+
+-   **The Problem:** OverlayGammaChart built x-axis labels from Session B's measurement array (`ptsB`) but plotted Session A's gamma values from a separate array (`ptsA`). Chart.js aligned data strictly by array index. When the two sessions used different grayscale ramp step counts (e.g., Session A = 21 steps, Session B = 11 steps), Session A's points were mapped to Session B's stimulus positions and extra Session A points were silently dropped. The Target 2.2 reference line also had only 11 points, further mis-representing the ideal curve.
+-   **The Solution:** Replaced positional index approach with union-label + keyed-lookup pattern matching OverlayDeChart:
+    -   Compute `allStimuli` as sorted union of stimulus_pct values from both sessions
+    -   Generate labels array from allStimuli (e.g., `['0%', '5%', '10%', ..., '100%']`)
+    -   Map each dataset using `find()` lookup by stimulus_pct with null fallback for missing points
+    -   Target 2.2 line uses same allStimuli array length for consistent x-axis
+-   **Impact on existing workflows/APIs:** No API changes. Only affects frontend chart rendering — both pre- and post-calibration passes now display correctly regardless of differing ramp step counts.
+
+---
+
+## 🧪 Testing & Validation
+
+### Environment Details
+
+-   **OS / Target Display:** macOS (CI) / Hisense U8G (production)
+-   **Hardware/Mock Used:** Simulated measurement arrays with differing ramp counts
+
+### Verification Steps
+
+1.  Run `npm run lint` — no errors
+2.  Run `npm run test:unit` — all 80 tests pass including new regression test
+3.  Verify new test renders 21 distinct x-axis tick labels when Session A has 21 steps and Session B has 11 steps
+
+### Firmware / TV Settings
+
+-    Verified against target display firmware version
+-    Local Dimming set to Medium during test runs
+-    Correct white point mode used (Warm1 for HDR, Warm2 for SDR)
+
+---
+
+## 📸 Visual Evidence (UI / Plot Changes)
+
+-   Chart now displays all measurement points from both sessions at correct stimulus positions
+-   No more silent data truncation when ramp counts differ
+
+---
+
+## 🧼 Checklist
+
+-    Code adheres to project style guidelines (formatting, typing, linting).
+-    Unit tests added/updated and passing.
+-    Documentation (README, inline docstrings) updated to reflect changes.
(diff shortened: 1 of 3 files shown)


jweber93:
Possible Improvements
Documentation: Add a docstring note to each setter — Briefly mention that the method is atomic with respect to concurrent deletes and other setters. This aids future maintainers.

Python
def set_code_scale(self, sid: str, code_scale: str) -> Dict[str, Any]:
    """Set code scale (8bit or 10bit).
    
    Atomic: held under SessionStore._lock to prevent TOCTOU races with delete/eviction.
    """
Consider extracting the lock wrapper into a helper — If more methods need the same pattern in the future, a decorator like @atomic_session_operation could reduce boilerplate. (Not necessary now, but consider for scale.)

Test the interaction with TTL eviction — The existing tests use explicit delete() calls. Adding a dedicated test for eviction-during-setter (with a controlled TTL mock) would strengthen confidence that the fix also protects against background eviction threads.

jweber93:
Incorporated all three items:

1. **Docstrings** — added atomicity note to each of the 6 setters and `repass()` documenting that they are held under SessionStore._lock to prevent TOCTOU races (#503).
2. **Decorator** — deferred per your note ("not necessary now"). The 7 methods are small enough that the inline `with self._lock:` is clearer than an indirection layer. Will revisit if the count grows.
3. **TTL eviction test** — added `test_setter_concurrent_with_ttl_eviction_no_resurrection`. Races `evict_expired_sessions()` against a setter with the session last_accessed_at pushed 30 days into the past. Verifies no crash and that the session is either fully updated or completely gone.

PR #511, opened 2026-06-30
Fix import routes blocking the event loop with sync disk I/O

### LINKED ISSUE #504: Bug: async import routes block the FastAPI event loop with synchronous, lock-holding disk I/O (server.py:1424-1455)
## Summary

`server.py` has exactly two `async def` route handlers — `import_zro_csv` and `import_generic_csv` — and both call straight-through synchronous, lock-holding, disk-writing code directly on the event loop instead of offloading it to a thread. Every other handler in `server.py` is a plain `def`, which FastAPI automatically runs in its threadpool; these two routes are the sole exceptions, so a large/slow import blocks the entire event loop — including SSE keep-alives and every other concurrent request — for the duration of the import.

## Evidence

`server.py:1424-1438`:

```python
@app.post("/api/session/{sid}/import/zro")
async def import_zro_csv(sid: str, file: UploadFile = File(...)):
    try:
        contents = await file.read()
    except Exception as exc:
        raise HTTPException(400, f"Could not read uploaded file: {exc}") from exc
    if len(contents) > _MAX_UPLOAD_BYTES:
        raise HTTPException(413, ...)
    session, import_meta = store.import_zro_bytes(sid, file.filename, contents)  # <-- sync, on the event loop
    _maybe_trigger_llm(sid, session)
    return {"session": _session_view(session), "import_summary": import_meta}
```

`import_generic_csv` (`:1441-1455`) is structurally identical. `store.import_zro_bytes` routes through `_locked_import` (`calibrator/session.py:2091-2113`), which holds `self._lock` (a `threading.RLock`) and calls `save_session()`, which does a blocking `path.write_text(...)`. Confirmed via `grep '^async def\|^def '` against `server.py` that these are the only two `async def` routes among the entire API surface — every other handler (including the other import-adjacent ones) is a plain synchronous `def`.

## Failure Scenario

A client uploads a CSV near the `_MAX_UPLOAD_BYTES` (10 MB, `server.py:32`) limit to `POST /api/session/{sid}/import/zro` while another client holds an open SSE stream (e.g. `GET /events/{sid}` or an in-flight LLM streaming response). The import's parse + lock acquisition + disk write all run synchronously inside the single-threaded asyncio event loop, so the SSE stream stalls and all other concurrent requests queue behind it until the import finishes — a multi-second freeze of the whole API for every connected client, not just the uploader.

## Root Cause

`async def` was adopted on these two routes only to support `await file.read()`; the rest of each handler body was left fully synchronous instead of being offloaded, which is a regression relative to the implicit threadpooling every other (plain `def`) handler in this file gets for free.

## Fix

Either (a) revert these two routes to plain `def` and read the upload via `file.file.read()` (matching every other handler's convention), or (b) keep `async def` and wrap the `store.import_zro_bytes` / `store.import_generic_bytes` calls in `await run_in_threadpool(...)` (Starlette's `starlette.concurrency.run_in_threadpool`).

## Test

A tight latency assertion would be timing-flaky in CI; the regression is best proven by code inspection / a structural test. Add a case to `tests/test_server_api.py` asserting that `import_zro_csv`/`import_generic_csv` are either plain `def` or, if `async def`, that the store calls are wrapped via `run_in_threadpool` (e.g. assert via `inspect` that the route function does not call `store.import_*_bytes` directly in its own frame).

## Affected File

- `server.py:1424-1455`

jweber93:
## 📺 Overview

Closes #504

### What does this PR do?
* **Component(s) affected:** Backend (`server.py`) — CSV import API routes

---

* **The Problem:** `import_zro_csv` and `import_generic_csv` were the only two `async def` route handlers in `server.py`. Every other handler is a plain `def`, which FastAPI automatically runs in its threadpool. These two routes called `store.import_zro_bytes` / `store.import_generic_bytes` directly inside the coroutine — synchronous, lock-holding (`threading.RLock`), disk-writing code (`save_session()` → `path.write_text(...)`) — which ran straight on the asyncio event loop. A large/slow import would block the entire event loop, stalling SSE keep-alives and every other concurrent request for the duration of the import.
* **The Solution:** Wrapped the `store.import_zro_bytes` / `store.import_generic_bytes` calls in `await run_in_threadpool(...)` (via `fastapi.concurrency.run_in_threadpool`, re-exporting Starlette's helper) so the blocking work runs off the event loop, consistent with how FastAPI already threadpools every plain `def` handler in the file.
* **Impact on existing workflows/APIs:** None — request/response shape and behavior are unchanged; only the execution context of the blocking call moved off the event loop.

---

### Verification Steps
1. Ran the full existing suite: `python -m pytest tests/test_server_api.py -q --no-cov` — all 146 tests pass.
2. Ran `ruff check server.py tests/test_server_api.py` — clean.
3. Added a structural regression test (`TestImportRoutesDoNotBlockEventLoop` in `tests/test_server_api.py`) asserting that if `import_zro_csv`/`import_generic_csv` remain `async def`, they must offload the store call via `run_in_threadpool` rather than calling `store.import_*_bytes(...)` directly in their own frame. A tight latency assertion would be flaky in CI, so this is verified via `inspect.getsource` instead.

---

## 🧼 Checklist
* [ ] Documentation (README, inline docstrings) updated to reflect changes. (not applicable — internal implementation detail)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xzy8Jajrew1GeV28z9dj5X)_

### DIFF
--- a/server.py
+++ b/server.py
@@ -1434,6 +1434,10 @@ async def import_zro_csv(sid: str, file: UploadFile = File(...)):
             413, f"File too large: {len(contents)} bytes (max {_MAX_UPLOAD_BYTES})"
         )
 
+    # Offload blocking, lock-holding disk I/O to the threadpool: this route is
+    # `async def` (for `await file.read()`), so unlike plain `def` handlers it
+    # is not threadpooled automatically and would otherwise stall the event
+    # loop for the duration of the import (#504).
     session, import_meta = await run_in_threadpool(
         store.import_zro_bytes, sid, file.filename, contents
     )
@@ -1453,6 +1457,8 @@ async def import_generic_csv(sid: str, file: UploadFile = File(...)):
             413, f"File too large: {len(contents)} bytes (max {_MAX_UPLOAD_BYTES})"
         )
 
+    # See import_zro_csv above: offload to threadpool so this async route
+    # doesn't block the event loop with synchronous, lock-holding disk I/O.
     session, import_meta = await run_in_threadpool(
         store.import_generic_bytes, sid, file.filename, contents
     )
--- a/server.py
+++ b/server.py
@@ -27,6 +27,7 @@ from urllib.parse import urlparse
 import httpx
 
 from fastapi import FastAPI, File, HTTPException, UploadFile
+from fastapi.concurrency import run_in_threadpool
 from fastapi.middleware.cors import CORSMiddleware
 
 _MAX_UPLOAD_BYTES = 10 * 1024 * 1024  # 10 MB limit for CSV uploads
@@ -1433,7 +1434,9 @@ async def import_zro_csv(sid: str, file: UploadFile = File(...)):
             413, f"File too large: {len(contents)} bytes (max {_MAX_UPLOAD_BYTES})"
         )
 
-    session, import_meta = store.import_zro_bytes(sid, file.filename, contents)
+    session, import_meta = await run_in_threadpool(
+        store.import_zro_bytes, sid, file.filename, contents
+    )
     _maybe_trigger_llm(sid, session)
     return {"session": _session_view(session), "import_summary": import_meta}
 
@@ -1450,7 +1453,9 @@ async def import_generic_csv(sid: str, file: UploadFile = File(...)):
             413, f"File too large: {len(contents)} bytes (max {_MAX_UPLOAD_BYTES})"
         )
 
-    session, import_meta = store.import_generic_bytes(sid, file.filename, contents)
+    session, import_meta = await run_in_threadpool(
+        store.import_generic_bytes, sid, file.filename, contents
+    )
     _maybe_trigger_llm(sid, session)
     return {"session": _session_view(session), "import_summary": import_meta}
 
--- a/tests/test_server_api.py
+++ b/tests/test_server_api.py
@@ -785,6 +785,38 @@ class TestZROImport:
         assert resp.status_code == 400
 
 
+class TestImportRoutesDoNotBlockEventLoop:
+    """Regression test for #504: the import routes must not run blocking,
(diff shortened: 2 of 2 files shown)


jweber93:
Possible Improvements
Consider documenting the pattern: Add an inline comment above both run_in_threadpool calls explaining why async routes with blocking I/O need explicit offloading. This helps future maintainers avoid reintroducing the same bug.
Example:

Python
# Offload blocking disk I/O to threadpool to avoid blocking the event loop
# (similar to how FastAPI threadpools plain `def` handlers automatically)
session, import_meta = await run_in_threadpool(
    store.import_zro_bytes, sid, file.filename, contents
)
Test both routes explicitly: The regression test is parametrized, which is good, but consider adding a separate integration test that actually uploads a large CSV and validates that concurrent requests (e.g., an SSE stream) are not blocked during the import. This would be a higher-level confirmation of the fix's effectiveness (though latency assertions should remain loose to avoid flakiness).

Audit other async routes: A quick grep to confirm no other async routes have similar blocking patterns would be prudent as a general best practice, though the issue description suggests these are the only two.

jweber93:
Thanks for the review!

- **Inline comment**: added in f2bf2a5 above both `run_in_threadpool` calls.
- **Concurrent-upload integration test**: skipping this one — the linked issue explicitly calls out that a latency-based concurrency assertion would be timing-flaky in CI and recommends the structural test instead, which is what's here. Happy to revisit if you'd rather have the flakier coverage.
- **Audit other async routes**: confirmed via `grep -n "^async def" server.py` — `import_zro_csv` and `import_generic_csv` are the only `async def` route handlers in the file (the other two matches, `lifespan` and `_session_cleanup_loop`, aren't routes), so no other instances of this pattern exist.

---

PR #514, opened 2026-07-01
Fix query_pass_decision TypeError on non-numeric measurement fields

### LINKED ISSUE #501: Bug: query_pass_decision raises TypeError on non-numeric measurement fields, producing HTTP 500 instead of degrading gracefully (llm.py:889-900, server.py:1961)
## Summary

`calcore/llm.py`'s `query_*` functions are required (per project convention, see AGENTS.md/QE_AUDIT.md) to **never raise** — malformed input must degrade to `None`, never propagate as a 500. `query_pass_decision()` violates this: it aggregates and calls `round()` on client-supplied measurement fields (`delta_e`, `effective_gamma`, `cct`) **before** entering its only `try:` block, so non-numeric values raise an uncaught `TypeError` that the `/api/session/{sid}/pass-decision` endpoint has no exception handling for.

## Evidence

`calcore/llm.py:889-900` (inside `query_pass_decision`, called for every measurement in the request body, well before the function's `try:` block which starts at `:1014` around the actual HTTP call):

```python
for m in measurements:
    label = (m.get("label") or "").strip()
    de = m.get("delta_e")
    if de is None:
        continue
    ...
    if is_color:
        colors.setdefault(color_name, []).append({
            "label": label,
            "dE": round(de, 2),   # <-- raises if de is not numeric
            ...
```

`server.py:1955-1990`:

```python
class PassDecisionReq(BaseModel):
    measurements: List[Dict[str, Any]]   # Pydantic does not validate dict value types
    repass_count: int = 0

@app.post("/api/session/{sid}/pass-decision")
def post_pass_decision(sid: str, req: PassDecisionReq):
    ...
    decision = _query_pass_decision(
        measurements=req.measurements,   # no try/except around this call
        ...
    )
```

I reproduced this directly: calling `query_pass_decision(measurements=[{"label": "White 50%", "delta_e": "bad"}], ...)` raises `TypeError: type str doesn't define __round__ method`. Via `TestClient` against the actual FastAPI app with `raise_server_exceptions=False`, `POST /api/session/{sid}/pass-decision` with this payload (LLM configured) returns **HTTP 500**.

## Failure Scenario

`POST /api/session/{sid}/pass-decision` with an LLM endpoint configured and body `{"measurements":[{"label":"White 50%","delta_e":"bad","stimulus_pct":50}]}` — any malformed/corrupted measurement payload from the frontend (e.g. a stale or partially-written record) crashes the adaptive pass-decision gate with a 500 instead of returning a typed `None`/fallback decision.

## Root Cause

The aggregation/`round()` calls on raw, untyped client-supplied dict values run outside the function's exception-handling boundary. Every other `query_*` function in `calcore/llm.py` only does arithmetic inside its own `try`/`except Exception: return None` block; `query_pass_decision` is the exception.

## Fix

Wrap the aggregation loop (lines ~889-960) in a `try`/`except (TypeError, ValueError): continue` per-measurement, or coerce each numeric field via a small `_safe_float(x)` helper before calling `round()`, mirroring the function's own graceful-degradation contract.

## Test

Add to `tests/test_adaptive_loop.py`: `test_non_numeric_delta_e_returns_none_not_raise`, asserting `query_pass_decision(measurements=[{"label": "White 50%", "delta_e": "bad"}], ...)` returns `None` rather than raising. Add a `TestClient` case in `tests/test_server_api.py` posting the same malformed body to `/api/session/{sid}/pass-decision` and asserting a 200 response (graceful fallback), not 500.

## Affected Files

- `calcore/llm.py:889-960` (`query_pass_decision`)
- `server.py:1960-1990` (`post_pass_decision`)

jweber93:
## 📺 Overview

Closes #501

### What does this PR do?
* **Component(s) affected:** `calcore/llm.py` (`query_pass_decision`), `/api/session/{sid}/pass-decision` endpoint in `server.py`

---

* **The Problem:** `query_pass_decision()` aggregated client-supplied measurement fields (`delta_e`, `effective_gamma`, `cct`) and called `round()`/arithmetic on them *before* entering its `try`/`except` block. A non-numeric value (e.g. `delta_e: "bad"`) raised an uncaught `TypeError`, which propagated through `post_pass_decision` in `server.py` as an HTTP 500 instead of degrading gracefully, violating the project convention that `query_*` functions never raise.
* **The Solution:** Added a `_safe_float()` helper that coerces `delta_e`, `effective_gamma`, and `cct` to `float`, returning `None` when a value isn't numeric. Measurements whose `delta_e` doesn't coerce are skipped (same as the existing "missing `delta_e`" path). Also added a check so that if no measurement has a usable `delta_e`, the function returns `None` immediately rather than making a pointless LLM call with empty data.
* **Impact on existing workflows/APIs:** None for well-formed requests. Malformed/corrupted measurement payloads now cause `/api/session/{sid}/pass-decision` to return its existing graceful fallback (`{"action": "accept", "reason": "LLM unavailable", ...}`, HTTP 200) instead of a 500.

* [x] Floating-point precision or data truncation risks have been addressed (non-numeric values are safely coerced/dropped rather than crashing).

---

### Environment Details
* N/A — this is a pure backend/API fix, no display hardware involved.

### Verification Steps
1. Run `python -m pytest tests/test_adaptive_loop.py tests/test_server_api.py -q`
2. Added `test_non_numeric_delta_e_returns_none_not_raise` and `test_non_numeric_effective_gamma_and_cct_do_not_raise` in `tests/test_adaptive_loop.py`, verifying `query_pass_decision` returns `None`/a valid decision instead of raising.
3. Added `test_non_numeric_delta_e_returns_200_not_500` in `tests/test_server_api.py`, posting a malformed measurement to `/api/session/{sid}/pass-decision` via `TestClient` and asserting HTTP 200.
4. Confirmed all three new tests fail with the original code (reproducing the bug) and pass with the fix. Full test suite (`python -m pytest -q`) passes.

### Firmware / TV Settings
* N/A

---

## 🧼 Checklist
* [ ] Documentation (README, inline docstrings) updated to reflect changes. (not needed — internal bug fix, no public API/behavior change)

---
_Generated by [Claude Code](https://claude.ai/code/session_01VXCppHZqJ9Z3AL9nfvjsat)_

### DIFF
--- a/calcore/llm.py
+++ b/calcore/llm.py
@@ -52,6 +52,16 @@ def _extract_json(text: str) -> str:
     return text[start:]
 
 
+def _safe_float(value: Any) -> Optional[float]:
+    """Coerce a client-supplied value to float, or None if it isn't numeric."""
+    if value is None:
+        return None
+    try:
+        return float(value)
+    except (TypeError, ValueError):
+        return None
+
+
 def _top_offenders(summary: Summary) -> Dict[str, Any]:
     """Extract worst-N patches by ΔE for compact LLM context (no raw XYZ)."""
     result: Dict[str, Any] = {}
@@ -888,7 +898,7 @@ def query_pass_decision(
     colors: Dict[str, List[Dict[str, Any]]] = {}
     for m in measurements:
         label = (m.get("label") or "").strip()
-        de = m.get("delta_e")
+        de = _safe_float(m.get("delta_e"))
         if de is None:
             continue
         stim_pct = m.get("stimulus_pct", 0)
@@ -908,12 +918,15 @@ def query_pass_decision(
                 "label": label,
                 "dE": round(de, 2),
                 "stimulus_pct": stim_pct,
-                "effective_gamma": m.get("effective_gamma"),
+                "effective_gamma": _safe_float(m.get("effective_gamma")),
                 "x": m.get("x"),
                 "y": m.get("y"),
                 "Y": m.get("Y"),
-                "cct": m.get("cct"),
+                "cct": _safe_float(m.get("cct")),
             })
+
+    if not grayscale and not colors:
+        return None
 
     # Compute per-region stats
     def _region_stats(items: List[Dict]) -> Optional[Dict]:
--- a/tests/test_adaptive_loop.py
+++ b/tests/test_adaptive_loop.py
@@ -262,6 +262,78 @@ class TestQueryPassDecision:
         assert result is not None
         assert result.action == "accept"  # falls back
 
+    def test_non_numeric_delta_e_returns_none_not_raise(self):
+        """Non-numeric delta_e values must degrade to None, never raise."""
+        llm = MagicMock()
+        llm.endpoint = "http://localhost:4000"
+        llm.model = "test-model"
+        llm.api_key = ""
+        llm.timeout = 30.0
+
+        result = query_pass_decision(
+            measurements=[{"label": "White 50%", "delta_e": "bad"}],
+            phase="pre_grayscale",
+            signal_range="full",
+            code_scale="8bit",
+            target_gamma=2.2,
+            target_peak_nits=120.0,
+            target_white_point=[0.3127, 0.3290],
+            target_gamut="bt709",
+            llm=llm,
+        )
+
+        assert result is None
+
+    def test_non_numeric_effective_gamma_and_cct_do_not_raise(self):
+        """Non-numeric effective_gamma/cct on an otherwise-valid measurement must not raise."""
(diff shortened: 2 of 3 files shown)


jweber93:
Minor notes (non-blocking)
stimulus_pct is not wrapped in _safe_float. It defaults to 0 via .get("stimulus_pct", 0), but a non-numeric value passes through as-is to the LLM prompt. Not critical — not used in _region_stats and just context text. Out of scope for this fix.
_region_stats iterates i["dE"] — safe because non-numeric delta_e measurements are excluded at the measurement level. No action needed.

PR #559, opened 2026-07-01
Autocal: loop abstractions, CMS controller, orchestrator, API + prefs (1a–1d, 1g)

jweber93:
## 📺 Overview

Closes #522, #523, #524, #525, #528 (sub-tasks 1a, 1b, 1c, 1d, 1g of #519).

Implements the backend half of the Autocal closed-loop measure → correct → apply → re-measure controller described in `docs/autocal-roadmap.md` (Item 1), scoped to the pieces that form one safe, self-contained, hardware-independent PR:

- **1a** — `MeasurementSource` / `ApplyTarget` interfaces with `ManualApplyTarget`, `AdbApplyTarget`, and a `CallableMeasurementSource` adapter.
- **1b** — A damped, oscillation-safe CMS correction controller with secant-method gain learning and per-control clamping.
- **1c** — The `AutocalLoop` orchestrator: per-primary measure → correct → apply → re-measure, stopping on `quality.py`'s ΔE gate or `max_iterations`, with cooperative cancellation.
- **1d** — `POST /api/session/{sid}/autocal/run`, `POST .../autocal/stop`, and an SSE stream at `GET .../autocal/stream`, following the existing LLM SSE broadcaster pattern.
- **1g** — `apply_mode` / `damping` / `max_iterations` preferences persisted in `.prefs.json`, with validation and README documentation.

**Deliberately excluded** (separate follow-up PRs, per the roadmap's own sequencing):
- **1e** (manual-apply frontend UX) — different layer/review surface; cleaner once this API shape is stable.
- **1f** (ADB auto-apply's fallback-to-manual UX) — the roadmap flags this as "Phase 2, once the manual loop exists," and its DoD requires verification against real Hisense U8G hardware, which this PR can't honestly provide.

### What does this PR do?
* **Type of change:** [x] New feature
* **Component(s) affected:** Backend — `calibrator/autocal*.py`, `server.py` (new endpoints + prefs)

---

## 🛠️ Technical Context & Implementation

* **The Problem:** Calibrating a TV's Color Management System today is manual guesswork per primary — the app can tell you a color is off but not run a convergent measure/adjust loop for you. A naive proportional controller mapping "raise Saturation by X" to measured ΔE oscillates or stalls because the TV's response is nonlinear and unknown a priori.
[...]

Issue #527, opened 2026-07-01
[Autocal 1f] ADB auto-apply mode


jweber93:
Part of #519. See `docs/autocal-roadmap.md` (Item 1). Phase 2 — layers onto the same controller once the manual loop exists.

**Spec:** Wire `ApplyTarget(ADB)` to `/api/adb/cms/set` with read-back verification.

**Definition of Done:** On a Hisense U8G, the loop applies without user interaction; verifies each write via get; falls back to manual mode on any ADB failure.

**Complexity:** Low–medium · **Recommended model:** Sonnet 5

jweber93:
Closed by #562 (merged). `FallbackApplyTarget(AdbApplyTarget, ManualApplyTarget)` wires ADB auto-apply with read-back verification (via the existing `AdbApplyTarget`) and falls back to a manual instruction — pausing the loop for user confirmation via the same `requires_user_action` gate — on any ADB failure. Hardware verification against a real U8G wasn't possible in this environment, per the roadmap's own note that this needs real-hardware testing; flagging that as a caveat rather than a blocker, consistent with how #559 handled the same limitation.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #519, opened 2026-07-01
Autocal: guided closed-loop measure → correct → apply → re-measure (CMS first)


jweber93:
## Summary

Add a Calman-style **autocal** controller that closes the calibration loop: trigger a measurement, import/analyze it, compute the correction, apply it, then re-measure — iterating per target until a quality gate is met or an iteration cap is hit.

The critical design choice is that the **apply step is pluggable**, so the loop is not tied to any one TV:

```
measure → compute correction → APPLY → re-measure → converge
                                 ├── manual: show "set Red Saturation → +3", user changes it on the TV, clicks "Remeasure"   (universal, phase 1)
                                 └── auto:   push via ADB                                                                    (Hisense U8G, phase 2)
```

The **manual-apply path is the highest-ROI starting point**: it works on *every* TV, needs no ADB / rooting / per-model integration, and is essentially the existing guidance engine wrapped in a tight measure/instruct/remeasure loop with convergence tracking. ADB auto-apply then layers onto the same controller.

Every primitive already exists; this issue is the orchestrator that chains them.

| Primitive | Where it lives today | Role |
|---|---|---|
| Trigger a measurement | `POST /api/bridge/measure` (ZRO Bridge) | measure |
| Import + analyze | watch folder → `_calcore_analyze()` → `Summary` | read result |
| Compute the adjustment | `calibrator/guidance.py` (CMS hue/sat/bright) | decide |
| Apply — manual | new instruction card + "Remeasure" button | turn the knob (any TV) |
| Apply — auto | `calibrator/adb_control.py` → `POST /api/adb/cms/set` | turn the knob (U8G) |
| Convergence gate | `calibrator/quality.py` (ΔE / CCT / gamma thresholds) | done? |

## User Story

As a user calibrating any TV, I want the app to measure, tell me exactly which CMS control to change (or apply it automatically if my TV supports ADB), and re-measure — looping until each primary is in tolerance — so I get Calman-style automation without manually tracking every adjustment.

## Design decisions

- **Pluggable apply.** `ApplyTarget` interface with `manual` (instruction + remeasure) and `adb` implementations. The controller never references ADB directly. **Manual is phase 1** and works everywhere; ADB is phase 2.
- **Pluggable measurement source.** The controller depends only on `MeasurementSource.measure(patch) -> XYZ`, so it runs over the ZRO Bridge today and over the ArgyllCMS backend (#520) or Integration Protocol backend (#521) later, unchanged.
[...]

PR #562, opened 2026-07-02
Autocal: manual-apply UX + ADB auto-apply fallback

### LINKED ISSUE #526: [Autocal 1e] Manual-apply UX: instruction card + Remeasure
Part of #519. See `docs/autocal-roadmap.md` (Item 1). **This is the universal, highest-ROI path — works on any TV.**

**Spec:** Instruction card ("set Red Saturation → +3"), a **Remeasure** button, and live per-primary ΔE progress.

**Definition of Done:** User can run a full manual loop end-to-end in the UI; clear "converged / keep going / out of range" states per primary; stop control present.

**Complexity:** Medium · **Recommended model:** Sonnet 5

jweber93:
## 📺 Overview

Closes #526, #527, and (since these were the last two open checklist items on the tracking issue) #519.

Implements the two remaining sub-tasks of the Autocal roadmap (Item 1, `docs/autocal-roadmap.md`), on top of the backend loop/API/prefs that already merged in #559:

- **1e — Manual-apply UX** (#526): a Color Tuner "Autocal" card that runs the guided measure → correct → apply → re-measure loop end-to-end in the UI.
- **1f — ADB auto-apply** (#527): `FallbackApplyTarget` wraps ADB apply with a manual-instruction fallback on any ADB failure.

### What does this PR do?
* **Type of change:** [x] New feature
* **Component(s) affected:** Backend (`calibrator/autocal_apply.py`, `calibrator/autocal_loop.py`, `server.py`), Frontend (new `AutocalCard`, wired into `ColorTuner.jsx`)

---

* **The Problem:** #559 shipped the autocal loop, controller, and API, but the loop's `_run_colour` would immediately re-measure after "applying" a manual correction — before the user had any chance to actually turn the knob on the TV. That's fine for real ADB writes (which take effect instantly) but silently broken for the manual path the roadmap calls "the universal, highest-ROI starting point": remeasuring a screen the user hasn't touched yet just reports a stale reading and (worse) lets the controller's tracked `current_value` drift out of sync with TV reality. There was also no frontend at all for the manual loop, and no fallback when an ADB write/read-back fails mid-run.
* **The Solution:** Add a pause-for-confirmation gate to the loop, a fallback-to-manual apply combinator, and a frontend card that drives both end-to-end.
  - `AutocalLoop` gains an opt-in `confirm_step`/`on_waiting` pause gate: whenever an iteration's apply reports `requires_user_action=True` (manual mode, or an ADB step that fell back to manual), the loop blocks until confirmed before re-measuring. Defaults to `None` (no pause), so existing callers/tests are unaffected.
  - `FallbackApplyTarget(primary, fallback)` in `autocal_apply.py`: tries the primary (ADB) apply; on failure, delegates to `ManualApplyTarget` and reuses the exact same `requires_user_action` signal to pause the loop. `server.py`'s `adb` apply mode now uses `FallbackApplyTarget(AdbApplyTarget, ManualApplyTarget)` instead of bare `AdbApplyTarget`.
  - `server.py`: new `POST /api/session/{sid}/autocal/confirm` endpoint + per-session `threading.Event` unblocks a paused loop; a stop request also unblocks it (checked each 1s poll). New `autocal_waiting` SSE event; `autocal_iteration` payloads now include `delta_e` and `requires_user_action` for the frontend to render live progress and the confirm gate.
  - New `AutocalCard` (frontend): apply-mode toggle (persists via existing `/api/prefs`), iteration cap, Start/Stop, a per-primary status grid (`Not started / Measuring… / Waiting on you / Converged / Keep going`) with live ΔE, and the "I made the change — Remeasure" confirm banner. Wired into `ColorTuner.jsx` above the existing Color Status card.
* **Impact on existing workflows/APIs:** Additive only — one new endpoint, two new SSE event/field additions (`autocal_waiting` event; `delta_e`/`requires_user_action` on `autocal_iteration`), one new frontend component. No existing endpoint or session shape changes. `AutocalLoop`'s new constructor params default to `None`, matching pre-existing behavior exactly when unset.

* [ ] Matrix transformations or LUT calculations have been validated. — N/A, no math changes (the controller from #559 is untouched).
* [x] Floating-point precision or data truncation risks have been addressed. — N/A, no new numeric computation.

---

### Environment Details
* **OS / Target Display:** N/A — backend logic tested against fakes/synthetic models and a mocked `EventSource` in the frontend; no physical TV used (ADB hardware verification is explicitly out of scope per the roadmap, same as #559).
* **Hardware/Mock Used:** `unittest.mock.patch` over `calibrator.adb_control` and `_SessionCmsMeasurementSource`, a synthetic per-colour display-response model, FastAPI `TestClient`, and a hand-rolled `MockEventSource` for the React component tests.

### Verification Steps
1. `python -m pytest tests/test_autocal_controller.py tests/test_autocal_apply.py tests/test_autocal_loop.py tests/test_autocal_api.py -q` — 66 tests, all passing (26 new).
2. `python -m pytest tests/ -q` — full suite, 1342 passed, 2 skipped, no regressions.
3. `ruff check` clean on all changed Python files.
4. `npx vitest run` (frontend) — 104 tests passing (8 new, in `AutocalCard.test.jsx`).
5. `npx eslint .` clean; `npm run build` succeeds.

### Firmware / TV Settings
* Not applicable — no physical TV was used for this PR.

---

<details>
<summary><b>Click to expand Visuals</b></summary>

### Before
No Autocal UI existed — the Color Tuner step only had the existing manual `ActionPlan` card with no loop/convergence tracking.

### After
A new "Autocal — guided measure / correct / apply loop" card at the top of Color Tuner: apply-mode toggle, iteration cap, Start/Stop, per-primary ΔE + status grid, and a confirm banner ("Raise Red Saturation to 4, then confirm" + "I made the change — Remeasure" button) that appears whenever the loop pauses for user action.

</details>

---

## 🧼 Checklist
* [x] Code adheres to project style guidelines (formatting, typing, linting) — `ruff check` and `eslint` both clean.
* [x] Unit tests added/updated and passing — 26 new backend tests, 8 new frontend tests, full suite green.
* [x] Documentation (README, inline docstrings) updated to reflect changes — expanded "Autocal (Guided CMS Loop)" section with the confirm endpoint, `autocal_waiting` event, and the new UI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SxLqGAEGSLmk43MMSraSN6)_


jweber93:
Actionable Feedback for PR #562
Status: ✅ Well-scoped, comprehensive PR ready for merge after addressing a few polish items.

Strengths
Clear Problem-Solution Alignment: The PR directly addresses the roadmap items (1e & 1f) with well-articulated design choices. The pause-for-confirmation gate elegantly solves the core issue: remeasuring before the user has acted leads to stale readings and controller drift.

Strong Test Coverage: 26 new backend tests + 8 frontend tests, with the full suite passing (1342 tests). Test coverage specifically validates the new pause/confirm flow and ADB fallback logic.

Backward Compatible: New params default to None, so existing callers work unchanged. The API is additive only (1 new endpoint, 2 SSE field additions, 1 new component).

Well-Documented: Clear docstrings in FallbackApplyTarget, AutocalLoop, and the updated README. The commit message and PR body are excellent.

Issues & Suggestions
1. Race Condition in confirm_step Polling

Python
def confirm_step() -> bool:
    while not loop.cancelled:
        if confirm_event.wait(timeout=1.0):
            confirm_event.clear()
            return True
    return False
Issue: If loop.cancel() is called between the while check and confirm_event.wait(), the loop could block for the full 1s timeout before exiting.
Suggestion: Check loop.cancelled again inside the loop after the wait returns, or use a combined cancel+confirm event pattern.

2. AutocalCard: Iteration State Not Reset Per Run

JavaScript
if (!running) setColourState(initialColourState(colours));
This only resets when transitioning into a not-running state, but running doesn't immediately become false in handleStart(). This could leave stale state if a run fails early.
Suggestion: Reset colourState before calling autocalRun(), not just when running changes.

3. Missing Error Handling in SSE Connection The es.onerror handler retries up to 5 times, but if the backend crashes mid-run, the frontend won't know the loop is dead—it'll just show "Autocal progress stream disconnected" after ~7.5s.
Suggestion:

Add a heartbeat/keep-alive event from the backend so the frontend knows the loop is still alive.
Or: Have the frontend check loop status (via a new /api/session/{sid}/autocal/status GET) when the stream drops, to distinguish "server went down" from "normal network blip."
4. Delta-E Display Precision in AutocalCard

JavaScript
<span dangerouslySetInnerHTML={{ __html: fmtDe(cs.deltaE) }} />
Using dangerouslySetInnerHTML for a simple number is risky (though fmtDe should be safe). Consider wrapping in a proper React component instead.

5. Confirm Button UX: No Visual Feedback While Confirming When the user clicks "I made the change — Remeasure," the button should show a loading state or disable temporarily to prevent double-clicks.
Suggestion:

JavaScript
<Button small primary onClick={handleConfirm} disabled={confirming}>
  {confirming ? 'Confirming…' : 'I made the change — Remeasure'}
</Button>
6. Missing TypeScript / JSDoc in AutocalCard The component has several inline magic values (retry delays, status map). Adding JSDoc comments would clarify the design intent.

Minor Code Polish
Line 2493 in server.py: The comment wraps oddly; consider breaking it differently for readability.
AutocalCard.jsx: The three-column grid uses className="three-col" but there's no CSS shown—ensure that class is defined in the stylesheet.
Tests: The test names in test_autocal_api.py (e.g., test_confirm_unblocks_manual_apply_pause) are verbose but clear; consider adding brief docstrings for maintainability.
Questions Before Merge
Measurement Caching: Does _SessionCmsMeasurementSource cache measurements? If the user confirms immediately but the TV hasn't actually changed yet, will the re-measurement still report stale data?

Stop During Wait: The comment says "Stop also unblocks," but the code checks loop.cancelled every 1s. What's the user experience—do they see "Waiting" briefly after clicking Stop, or does it respond immediately?

Fallback Message Clarity: When ADB fails and falls back, the message becomes "Auto-apply failed (ADB error); falling back to manual — Raise Hue to 5". Is this clear enough, or should it say "Try manually:" to signal the fallback more explicitly?

jweber93:
Thanks for the thorough review! Pushed a follow-up commit addressing the actionable items:

**Fixed**
- **#2 (state reset)** — real bug, though the actual mechanism was different than described: `ColorTuner.jsx` passes `colours={(cd.colors || []).map(c => c.name)}`, a fresh array literal on every render. The reset `useEffect` depended on that array *reference*, so any unrelated parent re-render (e.g. the session SSE stream ticking) while `running` was `false` would wipe converged/in-progress status back to "Not started" even though the colour selection hadn't changed. Fixed by keying the effect on `colours.join(',')` instead of the array reference. Added a regression test (`does not reset an in-progress colour when the parent passes a new colours array with the same content`) that fails on the old code.
- **#5 (confirm button feedback)** — added a `confirming` state; the button now disables and reads "Confirming…" while the request is in flight.
- Bonus: `autocal_done` now distinguishes "Autocal run stopped." (cancelled) from "Autocal run finished." (completed), which also answers the "Stop during wait" question below.

**Not changing, with reasoning**
- **#1 (confirm_step race)** — `while not loop.cancelled: if confirm_event.wait(timeout=1.0): ...` already re-checks `loop.cancelled` at the top of every iteration, including right after a timed-out wait — so the bound is exactly the documented "noticed within ~1s," not an unbounded block. No code path skips that check.
- **#4 (`dangerouslySetInnerHTML` for ΔE)** — matches the existing pattern already used one card up in the same file (`ColorTuner.jsx`'s Color Status card does the same with `fmtDe(m.delta_e)`), and `fmtDe` only ever formats a number via `toFixed(2)` — no untrusted input reaches it. Kept for consistency with the surrounding code rather than introducing a second convention.
- **#3 (SSE heartbeat / status endpoint)** — agreed this would be a nice hardening follow-up, but adding a new `/autocal/status` endpoint (or a heartbeat event) changes the API surface, which felt like it deserved its own PR rather than folding into this one silently. Happy to open a follow-up issue if you want it tracked.
- **#6 / minor polish** (JSDoc, line-2493 comment wrap, test docstrings) — left as-is; the project's existing convention (including neighboring test files) doesn't consistently JSDoc/docstring individual test cases, and I didn't find an awkwardly-wrapped comment at the current line 2493 (line numbers shifted a bit across commits) — happy to take a specific line pointer if I'm missing it.

**Answers**
- **Measurement caching:** No caching — `_SessionCmsMeasurementSource.measure()` always fires a real ZRO Bridge trigger and polls `session["cms_measurements"]` for a genuinely new reading each call. If the user confirms without actually changing the TV, the next measurement is a real, fresh reading of the unchanged picture — so it accurately reports "no improvement" rather than serving something stale, and the controller/loop just tries again on the next iteration.
- **Stop during wait:** Now addressed via the cancelled-vs-done messaging above. Latency-wise: `/autocal/stop` calls `loop.cancel()`, and the paused `confirm_step()` notices within its next `wait(timeout=1.0)` cycle — so worst case ~1s, often faster.
- **Fallback message clarity:** Current wording already says "falling back to manual" explicitly (e.g. `"Auto-apply failed (ADB error); falling back to manual — Raise Hue to 5"`), which I think conveys the same signal as "Try manually:" — left as-is, but easy to tweak if you'd still prefer that phrasing.

All tests green after the fix: 1342 backend / 111 frontend (3 new). `ruff` and `eslint` clean.

---

PR #563, opened 2026-07-03
Add ArgyllCMS spotread backend for license-free meter reads

jweber93:
## 📺 Overview

Adds an ArgyllCMS `spotread` measurement backend to the ZRO Bridge, so a supported colorimeter/spectrophotometer can be read directly — no ColourSpace/ZRO license needed. This PR covers the core, independently-testable slice of #520: the wrapper itself, done correctly (async, absolute-nit output) and fully tested.

Closes #530, #533, #534, #536 (sub-issues of #520, roadmap Item 2). Not included here (separate follow-up PRs): #531 (meter discovery/selection), #532 (CCMX/CCSS correction files), #535 (prefs UI + install docs).

### What does this PR do?
* **Type of change:** [ ] Bug fix | [x] New feature | [ ] Refactoring | [ ] CI/CD or Tooling
* **Component(s) affected:** ZRO Bridge (`tools/zro-bridge`) — new `argyll` backend

---

## 🛠️ Technical Context & Implementation

* **The Problem:** The ZRO Bridge only supports triggering ColourSpace (via UI automation or its paid Integration Protocol) to take a reading. There's no way to read a meter directly without ColourSpace.
* **The Solution:** New `tools/zro-bridge/backends/argyll_backend.py` wraps `spotread -e -O` for a single-shot emissive read, parses XYZ (or Yxy, converted to XYZ) out of its stdout, and classifies failures (`not_found`, `no_meter`, `timeout`, `parse_error`, `spotread_error`) into structured errors. The subprocess call is wrapped in an async `read_xyz_async()` that runs it via `fastapi.concurrency.run_in_threadpool`, matching the async-I/O pattern already used in `server.py`, so it never blocks the event loop. `bridge.py` gains a third `backend: "argyll"` option alongside `pyautogui`/`remote_control`, plus `argyll_spotread_path` / `argyll_port` / `argyll_timeout_s` config keys.
* **Impact on existing workflows/APIs:** `/status`, `/measure`, and `/measure/sequence` in `bridge.py` are now `async def` (previously sync `def`); the `pyautogui`/`remote_control` code paths are unchanged in behavior but now also run through `run_in_threadpool` for consistency, so no backend is left blocking the loop. Existing `pyautogui`/`remote_control` configs and behavior are unaffected.

### Mathematical / Data Integrity Checks (If Applicable)
* [x] Matrix transformations or LUT calculations have been validated. — N/A, no matrix/LUT math in this PR.
[...]

Issue #535, opened 2026-07-01
[Argyll 2f] Prefs + install/permissions docs


jweber93:
Part of #520. See `docs/autocal-roadmap.md` (Item 2).

**Spec:** Backend selection (`bridge`/`argyll`), meter type, and correction path in prefs; ArgyllCMS install + meter-permission docs (udev on Linux).

**Definition of Done:** Documented; persisted; graceful "ArgyllCMS not found" messaging.

**Complexity:** Low · **Recommended model:** Haiku 4.5 / local

Issue #520, opened 2026-07-01
ArgyllCMS direct-meter backend — license-free, end-to-end measurement (SDR + HDR)


jweber93:
## Summary

Add an **ArgyllCMS** measurement backend that reads the colorimeter/spectrophotometer directly (via `spotread`), removing ColourSpace ZRO from the measurement loop entirely. Combined with Dogegen (pattern generation) and ADB (apply), this makes the whole app consumable with **no paid products** — the app owns pattern → measure → compute → apply end to end.

Today the ZRO Bridge only *triggers* ZRO to take a reading and depends on ColourSpace (a paid product). ZRO's only role in the loop is reading the meter; [ArgyllCMS](https://www.argyllcms.com/) (free, open-source) talks to the same meters (i1Display, i1Pro, Klein K10-A, Colorimetry Research, etc.) and streams XYZ/xyY on demand.

## User Story

As a user who owns a supported colorimeter but not a ColourSpace license, I want the app to read my meter directly, so that I can run a complete SDR or HDR calibration using only free software.

## HDR support

Works cleanly — ArgyllCMS is EOTF-agnostic measurement transport:

- `spotread -e` (emissive) returns XYZ with **Y in absolute cd/m²** — exactly what PQ/ST.2084 targets need.
- HDR **pattern generation stays with Dogegen** (ST.2084 Rec.2020 10-bit); Argyll only replaces ZRO's meter-reading role.
- Analysis is already HDR-ready (`calcore/eotf.py`, `analysis.py` compute PQ error today).
- Caveat (true for any meter path): colorimeters need a per-panel spectral correction on modern wide-gamut panels (QD-OLED/WOLED/QD-LCD). Argyll supports this via CCMX/CCSS files (`spotread -X`). Expose a "meter correction file" field in the UI. Spectros (i1Pro) don't need it; colorimeters do.

## Implementation Checklist

- [ ] `tools/zro-bridge/backends/argyll_backend.py` (or a new `meter_backend.py`) wrapping `spotread` for single-shot emissive reads (`-e`, `-O` for one-shot, parse XYZ/xyY)
- [ ] Meter discovery (`spotread -?` / instrument list) and selection surfaced to the app
- [ ] Optional CCMX/CCSS correction-file path passed through to `spotread -X`
- [ ] Display refresh-sync / integration-time options for flicker-prone panels
- [ ] Return XYZ in the same shape the autocal `MeasurementSource` interface expects (issue for autocal orchestrator)
[...]

PR #566, opened 2026-07-03
Add ArgyllCMS meter discovery and selection

### LINKED ISSUE #531: [Argyll 2b] Meter discovery + selection
Part of #520. See `docs/autocal-roadmap.md` (Item 2).

**Spec:** Enumerate connected instruments and surface meter selection to the app.

**Definition of Done:** App lists detected meters; selection persists across sessions.

**Complexity:** Medium · **Recommended model:** Sonnet 5

jweber93:
## 📺 Overview

Adds ArgyllCMS meter discovery and selection to the ZRO Bridge's `argyll` backend (issue #520, roadmap Item 2), so the app can list connected colorimeters/spectrophotometers and remember which one to use.

Closes #531.

### What does this PR do?
* **Type of change:** [ ] Bug fix | [x] New feature | [ ] Refactoring | [ ] CI/CD or Tooling
* **Component(s) affected:** ZRO Bridge (`tools/zro-bridge`) — `argyll` backend; `server.py` — new proxy endpoints + prefs

---

* **The Problem:** #563 added the `argyll` backend's single-shot read (`spotread -e -O`), but there was no way to see which meters are actually connected, or to remember a choice across app restarts — every read silently used whatever `argyll_port` happened to be baked into `bridge.json`.
* **The Solution:** New `list_instruments()`/`list_instruments_async()` parse the detected-meter list out of `spotread -?`'s usage text, surfaced through a new bridge `GET /instruments` endpoint and a `POST /config/argyll-port` runtime selector; `server.py` proxies both and persists the choice in `.prefs.json`.
  - `argyll_backend.list_instruments()` runs `spotread -?` and parses the numbered `-c listno` instrument list out of its usage text (the standard way every Argyll CLI tool surfaces detected ports — there's no separate "list instruments" subcommand). An empty list is a valid "nothing connected" result, not an error.
  - `bridge.py` gains `GET /instruments` (backend=argyll only) and `POST /config/argyll-port`, which sets `_config["argyll_port"]` for the rest of the bridge process's life — an in-memory runtime override, consistent with the bridge's existing "simple config dict" architecture (no other backend setting is remotely mutable either; a restart falls back to `bridge.json`).
  - `server.py` proxies both (`GET /api/zro/bridge/instruments`, `POST /api/zro/bridge/instrument`), and persists the selection (`port` + `instrument_name`) in `.prefs.json` under a new `"argyll"` key — same pattern already used for `bridge_url`. On selection, it also best-effort pushes the port to the bridge immediately; if the bridge is unreachable the selection still saves and takes effect next time the app reconnects.
* **Impact on existing workflows/APIs:** Purely additive — no existing endpoint's behavior or contract changes. `_load_prefs()` now also loads the `"argyll"` key if present in an existing `.prefs.json` (harmless no-op for files written before this change).

* [x] N/A — no matrix/LUT/floating-point math in this PR (string parsing + config plumbing only).

---

## 🧪 Testing & Validation

- `tests/test_argyll_backend.py`: `list_instruments()`/`list_instruments_async()` parsing (multiple instruments, zero instruments, not-found, timeout), plus bridge-level integration tests for `GET /instruments` (backend gating, detected-list + `selected_port` in response) and `POST /config/argyll-port` (updates runtime config, and actually drives the next `/measure` call's spotread invocation).
- `tests/test_zro_bridge.py`: `server.py` proxy tests for `GET /api/zro/bridge/instruments` (not-configured, proxies bridge response, unreachable-bridge error) and `POST /api/zro/bridge/instrument` (persists to `_prefs`, pushes to bridge, degrades gracefully when the bridge is unreachable or unconfigured).

All `spotread`/bridge calls are mocked — no real ArgyllCMS installation or hardware meter required.

### Verification Steps
1. `pip install -r requirements.txt`
2. `python -m pytest tests/test_argyll_backend.py tests/test_zro_bridge.py -v` — 26 + 17 passed
3. `python -m pytest tests/` — full suite passes (~82% coverage, threshold 70%)
4. `ruff check tools/zro-bridge/backends/argyll_backend.py tools/zro-bridge/bridge.py server.py tests/test_argyll_backend.py tests/test_zro_bridge.py` — clean

---

## 🧼 Checklist
* [x] Code adheres to project style guidelines (formatting, typing, linting) — `ruff check` clean.
* [ ] Documentation (README, inline docstrings) updated to reflect changes. — A frontend picker UI and the fuller install/permissions docs are scoped to #535 (separate, already-tracked follow-up); this PR documents the new endpoints inline.

### DIFF
--- a/server.py
+++ b/server.py
@@ -422,6 +422,10 @@ _prefs: Dict[str, Any] = {
         "pattern_generator": "dogegen",
     },
     "autocal": dict(_AUTOCAL_DEFAULTS),
+    # Selected ArgyllCMS meter (issue #531) — port is the spotread -c listno
+    # index; instrument_name is the display string from the last discovery
+    # scan, kept alongside so the UI can show a selection without re-scanning.
+    "argyll": {"port": "", "instrument_name": ""},
 }
 
 
@@ -435,7 +439,7 @@ def _load_prefs() -> None:
     except Exception:
         logger.warning("Failed to parse .prefs.json; using defaults", exc_info=True)
         return
-    for key in ("dogegen", "bridge_url", "watch_folder", "llm", "session_defaults"):
+    for key in ("dogegen", "bridge_url", "watch_folder", "llm", "session_defaults", "argyll"):
         if key in saved:
             _prefs[key] = saved[key]
     if "autocal" in saved:
@@ -579,6 +583,11 @@ class TvSettingsReq(BaseModel):
 
 class ZroBridgeConfigBody(BaseModel):
     url: str
+
+
+class ZroBridgeInstrumentBody(BaseModel):
+    port: Optional[str] = None
+    instrument_name: Optional[str] = None
 
 
 class WatchConfigBody(BaseModel):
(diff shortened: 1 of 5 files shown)


jweber93:
Actionable Feedback
1. Consider documenting the selected_port response field
The GET /instruments endpoint includes "selected_port" in its response (the current argyll_port from config), but the PR description doesn't mention it. If this is meant to be part of the API contract, adding one sentence to the endpoint docstring (bridge.py L159) would clarify that it reflects the last selection or bridge.json default.

2. Verify .prefs.json migration path for existing users
The code safely handles .prefs.json files written before this change (the new "argyll" key is omitted = no-op on load). Confirm: Do you want to actively migrate old prefs files on first write, or is silent creation on first selection sufficient? Current behavior (silent creation) seems reasonable but worth a quick verification.

3. Test the race condition: concurrent /config/argyll-port + /measure
The implementation uses _config["argyll_port"] = port to update runtime state. test_argyll_backend.py L372 verifies the override takes effect on the next /measure, but there's no test for a concurrent scenario (e.g., POST /config/argyll-port from one thread while /measure is in-flight on another). The global _config dict is not locked in bridge.py — consider adding a test or a lock if needed.

jweber93:
Addressed in 89718db:

1. **`selected_port` docs** — added a sentence to `GET /instruments`' docstring: it reflects `bridge.json`'s static `argyll_port` until a `POST /config/argyll-port` call overrides it in memory for the rest of the process's life.
2. **Prefs migration** — no code change: this follows the exact same pattern already used for `bridge_url`, `llm`, `dogegen`, etc. — the key is simply absent from old `.prefs.json` files until the first write, which is a harmless no-op on load (`_load_prefs()` only copies keys that exist in the saved file). Actively migrating old files would be new machinery for one key where the existing convention already handles it; happy to build a real migration framework if you want one across the board, but didn't want to do it ad hoc for just this key.
3. **Concurrent write race** — added `test_config_write_concurrent_with_in_flight_measure_does_not_race`: fires `/measure` and, once it's already reading `_config`, a concurrent `POST /config/argyll-port` with a different value. No lock needed — `_config["argyll_port"] = port` is a single dict-key assignment, atomic under the GIL with no read-modify-write step, so there's no corruption hazard, only "which value does a concurrent read see" (answered: whatever was set when the read was dispatched, last-writer-wins for the next one). The test pins that down explicitly.

Full suite + ruff still clean.

---

PR #570, opened 2026-07-03
Guard PQ gamma plans against firmware tone-mapping knee

### LINKED ISSUE #548: [High] HDR gamma control plans drive per-point corrections into the firmware tone-mapping wall (no knee guard)
**From the QE audit** (`QE_AUDIT_FINDINGS.md`, PR #542). Severity: **High** · Confidence: **High** for the code path (no guard exists); the wall behavior is the audit's stated ground-truth invariant.

## Location
- `calibrator/guidance.py:403-448` (`u8g_gamma_control_plan`; also `preset_gamma_control_plan`, `gamma_recommendations`)

## Failure scenario
HDR10 U8G session, gamma step. At 80% stimulus the absolute PQ reference is **1555 nits** — above any real panel — so `eotf_from_luminance` returns > 1.1 *always* on a correctly tone-mapping display, and the plan emits **"Raise 80% by 5 — Measured gamma is too high here — brighten this point"** on every pass, forever. Same at 70% (778-nit reference) on ≤ ~750-nit panels. The operator ends up chasing firmware tone mapping with menu controls — exactly what the `QE_AUDIT.md` invariant forbids ("Do not 'correct' the 40–70% firmware tone-mapping wall via menu controls").

Worked example: 700-nit panel at 80% → `pq_inverse_eotf(700) ≈ 0.7399` → `log(0.7399)/log(0.8) = 1.35` → delta 0.35 > 0.10 → "up" recommendation every time.

## Evidence
Traced `session_view` → `u8g_gamma_control_plan(latest_pass, …, eotf="PQ (ST.2084)")`; grep for knee/wall/tone-map in `guidance.py` finds only docstrings. `tests/test_hdr_gamma_guidance.py` only tests uniform-scale cases, never a clipped peak.

## Root cause
Per-point plans compare absolute-PQ tracking with no notion of the panel's tone-map knee, so points whose PQ reference exceeds panel capability are treated as correctable errors.

## Fix
For PQ sessions, compute the knee (`pq_reference(n) ≥ panel_peak × margin`) and emit "hold — firmware tone-mapping region" for those points instead of up/down deltas. Must not suppress genuine midtone errors below the knee.

## Model tier
**4 — Frontier** (tone-mapping + PQ reasoning).

## Regression test
`tests/test_hdr_gamma_guidance.py` new case: 700-nit measured peak, correctly tone-mapped 70/80% points ⇒ direction "hold", not "up".

jweber93:
## 📺 Overview

HDR gamma control plans drove per-point corrections into the firmware tone-mapping wall, emitting "Raise 80% by 5" on every pass forever for PQ points whose ST.2084 reference exceeds panel peak.

Closes [#548](https://github.com/jweber93/tv-calibration/issues/548)

### What does this PR do?

-   **Component(s) affected:** calibrator/utils.py, calibrator/guidance.py, calibrator/__init__.py, tests/test_hdr_gamma_guidance.py

---

-   **The Problem:** `u8g_gamma_control_plan`, `preset_gamma_control_plan`, and `gamma_recommendations` compared per-point PQ tracking against the absolute ST.2084 reference with no notion of the panel tone-map knee. ST.2084 is an absolute EOTF: at 80% stimulus the reference is ~1555 nits and at 70% ~621 nits. On a sub-peak panel (e.g. 700 nits) the measured luminance always reads as "too dark" vs the unreachable reference, so `eotf_from_luminance` returns > 1.0 and the plan emits "Raise 80% by 5 — brighten this point" on every pass, forever. The operator ends up chasing firmware tone mapping with menu controls — exactly what `QE_AUDIT.md` forbids ("Do not correct the 40–70% firmware tone-mapping wall via menu controls").
-   **The Solution:** Add `pq_point_above_knee(stimulus_pct, peak_nits)` in `calibrator/utils.py` — true when `pq_eotf(stimulus) >= peak_nits * margin` (margin 1.0, a hard criterion: the reference is literally unreachable). Then:
    - `u8g_gamma_control_plan`: for PQ sessions, above-knee points emit `direction: "hold"` with reason "PQ reference luminance exceeds panel peak; firmware tone mapping is active here. Do not correct via menu controls." instead of up/down deltas.
    - `gamma_recommendations` / `preset_gamma_control_plan`: exclude above-knee points from the average and the low/high regional lists so overall/preset advice isn't dragged toward "too dark" by an unreachable reference. When all measured points are above the knee, emit a tone-mapping message instead of a misleading "take another reading" fallback.
-   **Impact on existing workflows/APIs:** No API changes. The per-point plan gains a `delta: null` field for knee-held points (previously always a number); consumers reading `direction` are unaffected. Below-knee points keep normal treatment so genuine midtone errors are still surfaced.

---

### Environment Details

-   **OS / Target Display:** macOS (CI) / Hisense U8G (production)
-   **Hardware/Mock Used:** Simulated PQ measurement arrays with hard-clip tone mapping

### Verification Steps

1.  `ruff check calibrator tests` — passes
2.  `pytest tests/test_hdr_gamma_guidance.py tests/test_session.py tests/test_utils.py tests/test_eotf_bt1886_degenerate.py` — 54 passed
3.  New regression cases (issue #548):
    - `TestU8gControlPlanPqKnee.test_sub_peak_panel_holds_above_knee_points` — 500-nit panel, correctly tone-mapped 70/80% ⇒ `hold` (not `up`), reason mentions tone mapping
    - `TestU8gControlPlanPqKnee.test_below_knee_too_dark_still_corrected` — below-knee half-scale points still get `up`
    - `TestU8gControlPlanPqKnee.test_knee_helper_boundary` — knee helper boundary + degenerate inputs
    - `TestPqKneeAggregate.test_recommendations_no_highlight_rec_for_clipped_peak` — "raise 70%/80%" rec suppressed for clipped peak
    - `TestPqKneeAggregate.test_preset_plan_holds_for_clipped_peak` — preset plan holds for clipped peak
    - `TestPqKneeAggregate.test_recommendations_all_above_knee_message` — tone-mapping message when all points above knee
4.  Updated `test_pq_too_dark_raises_points` (encoded the old buggy "all up" expectation) — below-knee 20/40/60% still `up`, 80% now `hold`.

### Firmware / TV Settings

-   Verified against target display firmware behavior (PQ tone-mapping knee)
-   Local Dimming set to Medium during calibration sessions
-   PQ (ST.2084) EOTF tracking invariant preserved (1.0 = perfect)

---

-   N/A — backend guidance logic only. Operators will see "Hold 80% — firmware tone-mapping region" in the gamma control plan instead of an infinite "Raise 80% by 5" loop.

---

## 🧼 Checklist

-    Code adheres to project style guidelines (formatting, typing, linting).
-    Unit tests added/updated and passing.
-    Documentation (inline docstrings) updated to reflect the knee guard and `PQ_KNEE_MARGIN` invariant.
-    No credentials, local absolute paths, or hardware-specific hardcoding leaked.

### DIFF
--- a/calibrator/guidance.py
+++ b/calibrator/guidance.py
@@ -12,7 +12,13 @@ from calcore.models import (
     Measurement,
 )
 from .profiles import TVProfile
-from .utils import delta_xy, eotf_from_luminance, is_pq_eotf, stimulus_pct_from_code_value
+from .utils import (
+    delta_xy,
+    eotf_from_luminance,
+    is_pq_eotf,
+    pq_point_above_knee,
+    stimulus_pct_from_code_value,
+)
 
 GAMMA_TRACKING_LEVELS = (20, 40, 60, 80)
 FINE_GAMMA_TRACKING_LEVELS = tuple(range(5, 100, 5))
@@ -341,10 +347,22 @@ def gamma_recommendations(
         stim_pct = gamma_nominal_pct(m, signal_range, code_scale, levels)
         if stim_pct is None:
             continue
+        # Skip points whose PQ reference exceeds panel peak — firmware tone
+        # mapping is active there and they would skew the aggregate advice
+        # toward "too dark" forever (issue #548). Below-knee points keep normal
+        # treatment so genuine midtone errors are still surfaced.
+        if pq and pq_point_above_knee(stim_pct, peak_nits):
+            continue
         eff_gamma = eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)
         if eff_gamma is not None:
             gammas.append({"stimulus_pct": stim_pct, "effective_gamma": eff_gamma})
     if not gammas:
+        if pq:
+            return [
+                "All measured gamma points sit in the firmware tone-mapping region "
+                "(PQ reference exceeds panel peak). Do not chase them with menu controls; "
+                "re-measure with lower-stimulus patches or raise the panel's HDR peak."
+            ]
         return ["Take another gamma reading. A valid effective gamma was not computed from the last pass."]
 
     avg = sum(m["effective_gamma"] for m in gammas) / len(gammas)
@@ -410,17 +428,41 @@ def u8g_gamma_control_plan(
     eotf: str = "gamma",
 ) -> List[Dict[str, Any]]:
     # PQ tracks against the ST.2084 reference (1.0 == perfect), not target_gamma.
-    effective_target = 1.0 if is_pq_eotf(eotf) else target_gamma
+    pq = is_pq_eotf(eotf)
+    effective_target = 1.0 if pq else target_gamma
     plan: List[Dict[str, Any]] = []
     for m in measurements:
         stim_pct = gamma_nominal_pct(m, signal_range, code_scale, levels)
         if stim_pct is None:
             continue
+        stim_label = f"{round(stim_pct):.0f}%"
+        # PQ points whose ST.2084 reference exceeds panel peak are inside the
+        # firmware tone-mapping region. The measured luminance will always read
+        # as "too dark" relative to an unreachable reference, so the plan would
+        # emit "raise this point" on every pass forever. Hold them instead and
+        # let the operator's midtone corrections reach the knee from below
+        # (issue #548 / QE_AUDIT.md tone-mapping invariant).
+        if pq and pq_point_above_knee(stim_pct, peak_nits):
+            eff_gamma = eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)
+            plan.append(
+                {
+                    "control": stim_label,
+                    "effective_gamma": round(eff_gamma, 3) if eff_gamma is not None else None,
+                    "delta": None,
+                    "direction": "hold",
+                    "amount": 0,
+                    "summary": f"Hold {stim_label} — firmware tone-mapping region",
+                    "reason": (
+                        "PQ reference luminance exceeds panel peak; firmware tone "
+                        "mapping is active here. Do not correct via menu controls."
+                    ),
+                }
+            )
+            continue
         eff_gamma = eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)
         if eff_gamma is None:
             continue
         delta = eff_gamma - effective_target
-        stim_label = f"{round(stim_pct):.0f}%"
         if abs(delta) < 0.10:
             direction, amount, summary = "hold", 0, f"Leave {stim_label} alone"
             reason = "This gamma point is already close enough to target."
@@ -483,10 +525,29 @@ def preset_gamma_control_plan(
         stim_pct = gamma_nominal_pct(m, signal_range, code_scale, levels)
         if stim_pct is None:
             continue
+        # Exclude firmware tone-mapping points so the preset recommendation
+        # isn't dragged toward "too dark" by an unreachable PQ reference
+        # (issue #548).
+        if pq and pq_point_above_knee(stim_pct, peak_nits):
+            continue
         eff = eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)
         if eff is not None:
             gammas.append(eff)
     if not gammas:
+        if pq:
+            return [
+                {
+                    "control": "Gamma preset",
+                    "direction": "hold",
+                    "amount": 0,
+                    "summary": "Keep current HDR tracking preset",
+                    "reason": (
+                        "All measured points sit in the firmware tone-mapping region "
+                        "(PQ reference exceeds panel peak). Preset changes won't recover "
+                        "an unreachable reference; raise the panel's HDR peak instead."
+                    ),
+                }
+            ]
         return []
     avg = sum(gammas) / len(gammas)
     delta = avg - effective_target
--- a/calibrator/guidance.py
+++ b/calibrator/guidance.py
@@ -343,6 +343,7 @@ def gamma_recommendations(
         return recs
 
     gammas = []
+    saw_above_knee = False
     for m in measurements:
         stim_pct = gamma_nominal_pct(m, signal_range, code_scale, levels)
         if stim_pct is None:
@@ -352,12 +353,17 @@ def gamma_recommendations(
         # toward "too dark" forever (issue #548). Below-knee points keep normal
         # treatment so genuine midtone errors are still surfaced.
         if pq and pq_point_above_knee(stim_pct, peak_nits):
+            saw_above_knee = True
             continue
         eff_gamma = eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)
         if eff_gamma is not None:
             gammas.append({"stimulus_pct": stim_pct, "effective_gamma": eff_gamma})
     if not gammas:
-        if pq:
+        # Only attribute the empty result to the tone-mapping region when we
+        # actually saw above-knee points; otherwise the readings themselves
+        # failed and the operator should re-measure rather than be told the
+        # wall is the cause.
+        if pq and saw_above_knee:
             return [
                 "All measured gamma points sit in the firmware tone-mapping region "
                 "(PQ reference exceeds panel peak). Do not chase them with menu controls; "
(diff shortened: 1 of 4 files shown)


calibrator/guidance.py
```diff
@@ -410,17 +428,41 @@ def u8g_gamma_control_plan(
...
+                    "direction": "hold",
+                    "amount": 0,
+                    "summary": f"Hold {stim_label} — firmware tone-mapping region",
+                    "reason": (
+                        "PQ reference luminance exceeds panel peak; firmware tone "
+                        "mapping is active here. Do not correct via menu controls."
+                    ),
+                }
+            )
+            continue
```
jweber93, line 472: **Bug:** this branch computes `eff_gamma` but never checks it for `None` before appending, unlike the non-knee branch right below (`if eff_gamma is None: continue`).

`Measurement.Y` defaults to `0.0`, and `eotf_from_luminance` returns `None` whenever `measured_nits <= 0`. So a failed/missing reading at an above-knee stimulus (e.g. 80% on a sub-peak panel) now produces `{"direction": "hold", "effective_gamma": None, ...}` with the reason "firmware tone mapping is active here" — falsely telling the operator this point was confirmed fine, when the measurement never actually registered. Before this PR, such a point would have been omitted entirely from the plan via the shared `continue` — not misleading, if imperfect.

Suggest hoisting the single `eotf_from_luminance(...)` call above the `if pq and pq_point_above_knee(...)` branch so both paths share the same `if eff_gamma is None: continue` guard — this also removes the duplicate call.

No existing test covers `Y == 0` (or another `None`-eff_gamma case) at an above-knee point, so this slipped through.

---

calibrator/utils.py
```diff
@@ -65,6 +65,34 @@ def is_pq_eotf(eotf: str) -> bool:
...
+    ST.2084 is an absolute EOTF: the encoded signal maps directly to nits, not
+    to a fraction of the display's peak. When the reference for a stimulus
+    exceeds what the panel can actually produce, firmware tone mapping takes
+    over and the measured luminance will always read as "too dark" relative to
+    that reference. Such points must not be "corrected" via menu controls — the
+    operator would chase the tone-mapping wall forever (issue #548).
+    """
+    if peak_nits <= 0 or stimulus_pct <= 0 or stimulus_pct >= 100:
+        return False
+    return pq_eotf(stimulus_pct / 100.0) >= peak_nits * margin
```
jweber93: **Reuse:** `calcore.eotf.pq_target_nits(n, peak_nits)` already computes `min(pq_eotf(n), peak_nits)` — the exact "is the PQ reference clipped at panel peak" concept this function needs — and is already used elsewhere for this purpose (`calibrator/session.py`, `calcore/analysis.py`, `calcore/targets.py`).

`pq_eotf(stimulus_pct / 100.0) >= peak_nits * margin` is mathematically equivalent to checking whether `pq_target_nits(...)` was clipped, but this reimplements it from scratch with a new import instead of reusing the existing, already-tested helper. Worth consolidating so the clipping math doesn't drift between two independent implementations over time.

---

jweber93 (commented):
Nice fix for the tone-mapping-knee loop (#548) — the exclusion logic and test coverage for the documented scenarios (500/700/1000-nit panels, boundary cases) look solid. Two inline notes: a correctness gap where a failed/zero reading at an above-knee point gets reported as a confirmed "hold" instead of being skipped, and a reuse opportunity around `calcore.eotf.pq_target_nits` which already does this clipping check elsewhere in the codebase.

---

jweber93:
## Self-review of final diff (f21bce8)

I did a full re-review of the diff after addressing the two inline notes from the first review. Findings below.

---

**Finding:**
- Severity: medium
- File: `calibrator/guidance.py` (`gamma_recommendations` and `preset_gamma_control_plan` empty-gammas fallback)
- Problem: The same above-knee / failed-reading conflation flagged on `u8g_gamma_control_plan` also existed in the aggregate fallback messages. When `gammas` ended up empty, the PQ branch unconditionally returned "All measured gamma points sit in the firmware tone-mapping region…" — but the emptiness could be caused by failed readings (`Y <= 0` → `eff_gamma is None`) at *below-knee* points, where no tone-mapping wall is involved. The operator would be wrongly told firmware tone mapping is the cause when actually the readings failed.
- Why it matters: Misdiagnoses a measurement failure as a hardware limit, sending the operator to raise the panel HDR peak instead of re-measuring.
- Suggested fix: Track `saw_above_knee` and only emit the tone-mapping message when an above-knee point was actually seen; otherwise fall back to "Take another gamma reading." / `[]`. **Applied in f21bce8.**

---

**Items verified clean (no action):**
- `delta: null` on knee-held plan entries — confirmed `frontend/src/components/ActionPlan.jsx` only reads `direction`/`amount`/`summary`/`reason`/`priority`, not `delta` or `effective_gamma`, so the new null is safe for the UI.
- `pq_target_nits(n, peak) < pq_eotf(n)` vs the original `pq_eotf(n) >= peak * margin`: the boundary flips at the exact `pq_eotf(n) == peak` point, but a perfect measurement there yields `eff_gamma == 1.0` → `hold` via the normal path, so the operator outcome is unchanged.
- The `low`/`high` regional lists in `gamma_recommendations` iterate `gammas`, which now excludes above-knee points, so the "raise 70%/80%" recommendation can no longer fire from a clipped peak.
- SDR path is untouched (`pq` flag gates every knee branch); `test_sdr_behaviour_unchanged` still passes.

**Test coverage added in f21bce8:** `test_above_knee_failed_reading_omitted`, `test_recommendations_failed_readings_not_blamed_on_tone_mapping`, `test_preset_plan_failed_readings_return_empty`. 57 passed across the relevant suites; `ruff check` clean.

Beyond the one finding above (now fixed), no further actionable issues found in the final diff.

calibrator/guidance.py
```diff
@@ -337,14 +343,32 @@ def gamma_recommendations(
...
         eff_gamma = eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)
         if eff_gamma is not None:
             gammas.append({"stimulus_pct": stim_pct, "effective_gamma": eff_gamma})
     if not gammas:
+        # Only attribute the empty result to the tone-mapping region when we
+        # actually saw above-knee points; otherwise the readings themselves
+        # failed and the operator should re-measure rather than be told the
+        # wall is the cause.
+        if pq and saw_above_knee:
```
jweber93, line 366: **Bug (partial fix):** `saw_above_knee` is set the moment *any* point is above-knee, but the message it gates says "**All** measured gamma points sit in the firmware tone-mapping region." Those aren't the same condition when the empty `gammas` result is actually caused by a mix of causes.

Concrete repro: 1000-nit panel, levels `(20, 40, 60, 80)`. 20/40/60% all have `Y = 0.0` (failed/missing readings — e.g. a photometer glitch), 80% is genuinely above the knee (`pq_eotf(0.8) ≈ 1555 > 1000`).
- 20/40/60%: not above-knee, but `eff_gamma` is `None` (measured_nits `<= 0`) → excluded from `gammas` without setting `saw_above_knee`.
- 80%: above-knee → `saw_above_knee = True`, excluded via `continue`.
- `gammas` ends up empty, `saw_above_knee` is `True` → returns "All measured gamma points sit in the firmware tone-mapping region... re-measure with lower-stimulus patches or raise the panel's HDR peak."

But 3 of the 4 points failed for an unrelated reason (no reading at all), not tone mapping — only the 80% point is legitimately above-knee. The message overclaims "All" and sends the operator toward "raise the panel's HDR peak," which won't fix a instrumentation/reading failure at 20/40/60%.

Same pattern applies to `preset_gamma_control_plan`'s identical `saw_above_knee` gate a few lines down (~547).

Suggest also tracking whether any *below-knee* point was seen (regardless of whether its `eff_gamma` came back `None`), and only emitting the tone-mapping message when there were zero below-knee points — i.e. every stimulus level that had a nominal match was genuinely above the knee, not just "at least one was."

---

jweber93 (commented):
Re-reviewed f21bce8. Both prior findings are fixed and verified:

- `u8g_gamma_control_plan` now computes `eff_gamma` once and checks `if eff_gamma is None: continue` before the knee branch, so a failed reading is omitted instead of reported as a false "hold." New `test_above_knee_failed_reading_omitted` covers it.
- `pq_point_above_knee` now delegates to `calcore.eotf.pq_target_nits` instead of reimplementing the clip check. The strict-`<` vs. the old `>=` boundary flip at `pq_eotf(n) == peak_nits` doesn't change operator-visible behavior, confirmed: a perfect reading there yields `eff_gamma == 1.0` → hold via the normal path either way.

One new issue in the self-review's own fix (f21bce8), flagged inline: the `saw_above_knee` guard added to `gamma_recommendations`/`preset_gamma_control_plan` fixes the pure "all failed, none above knee" and "all above knee" cases (both now have tests), but not the mixed case — it fires on "at least one above-knee point," while the message claims "**all** measured points" are in the tone-mapping region. A panel where most points failed to read and only one happens to be above-knee will get the wrong root-cause message.

---

PR #571, opened 2026-07-03
Fix mixed above-knee/failed-reading misattribution in gamma aggregate fallbacks

jweber93:
## 📺 Overview

Follow-up to #570 (issue #548). `gamma_recommendations` and `preset_gamma_control_plan` can misattribute an empty aggregate result to the firmware tone-mapping wall when the real cause is unrelated failed readings.

### What does this PR do?
* **Component(s) affected:** calibrator/guidance.py, tests/test_hdr_gamma_guidance.py

---

* **The Problem:** #570 added a `saw_above_knee` flag to `gamma_recommendations`/`preset_gamma_control_plan` so the empty-`gammas` fallback could tell "all points are in the tone-mapping region" apart from "readings failed." But the flag is set the moment *any* point is above the knee, while the fallback message claims "**All** measured gamma points sit in the firmware tone-mapping region." On a panel where some below-knee points simply failed to read (`Y <= 0` → `eff_gamma is None`) while a separate point happens to be genuinely above the knee, the message fires and wrongly tells the operator to "raise the panel's HDR peak" instead of re-measuring the points that actually failed.
* **The Solution:** Track a second flag, `saw_below_knee`, for any point that wasn't excluded by the knee guard (regardless of whether its `eff_gamma` came back `None`). The tone-mapping message is now only returned when `saw_above_knee and not saw_below_knee` — i.e. every measured point was genuinely above the knee, not just at least one.
* **Impact on existing workflows/APIs:** No API changes. The pure "all above knee" and "all failed, none above knee" cases from #570 are unaffected (already covered by existing tests); this only changes the previously-untested mixed case to return the generic "take another reading" fallback instead of the tone-mapping message.

---

### Environment Details
* **OS / Target Display:** CI / N/A (backend guidance logic only)
* **Hardware/Mock Used:** Simulated PQ measurement arrays

### Verification Steps
1. `ruff check calibrator tests` — passes
2. `pytest tests/test_hdr_gamma_guidance.py tests/test_session.py tests/test_utils.py tests/test_eotf_bt1886_degenerate.py` — all passing
3. New regression cases:
   - `test_recommendations_mixed_failure_and_knee_not_blamed_on_tone_mapping` — 1000-nit panel, 20/40/60% failed readings + genuinely above-knee 80% ⇒ "take another gamma reading," not the tone-mapping message
   - `test_preset_plan_mixed_failure_and_knee_returns_empty` — same mixed scenario ⇒ preset plan returns `[]`, not the tone-mapping hold

---

## 🧼 Checklist
* [x] Documentation (inline comments) updated to reflect the `saw_below_knee` invariant.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KJv3BfWDpY13iKaw7ThUSp)_

### DIFF
--- a/calibrator/guidance.py
+++ b/calibrator/guidance.py
@@ -356,6 +356,8 @@ def gamma_recommendations(
         if pq and pq_point_above_knee(stim_pct, peak_nits):
             saw_above_knee = True
             continue
+        # Mark that we saw at least one non-above-knee point, regardless of
+        # whether its reading succeeded.
         saw_below_knee = True
         eff_gamma = eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)
         if eff_gamma is not None:
@@ -544,6 +546,8 @@ def preset_gamma_control_plan(
         if pq and pq_point_above_knee(stim_pct, peak_nits):
             saw_above_knee = True
             continue
+        # Mark that we saw at least one non-above-knee point, regardless of
+        # whether its reading succeeded.
         saw_below_knee = True
         eff = eotf_from_luminance(m.Y, peak_nits, stim_pct, eotf)
         if eff is not None:
--- a/calibrator/guidance.py
+++ b/calibrator/guidance.py
@@ -344,6 +344,7 @@ def gamma_recommendations(
 
     gammas = []
     saw_above_knee = False
+    saw_below_knee = False
     for m in measurements:
         stim_pct = gamma_nominal_pct(m, signal_range, code_scale, levels)
         if stim_pct is None:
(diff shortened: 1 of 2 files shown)


jweber93:
Recommendations Before Merging
Consider Adding an Explicit "None" Flag Comment (Minor Polish)

Line 359 sets saw_below_knee = True unconditionally when a measurement passes the knee check. This is correct, but a comment like # Mark that we saw at least one non-above-knee point, regardless of whether its reading succeeded would make the intent clearer for future maintainers.
Verify Test Coverage for Edge Cases (Pre-Merge Sanity Check)

✅ Covered: Mixed failure + above-knee
✅ Covered: All above-knee (existing test)
✅ Covered: All failed, none above-knee (existing test)
Consider: Single measurement above-knee? Single measurement failed? (Likely already covered in upstream tests, but worth confirming if you haven't.)
Update Issue/PR Cross-Reference (Meta)

The PR references #570 and #548. Confirm those are linked in the body for traceability. (Already done — good job.)
Consider a Broader Rename (Future Refactor, Not Blocking)

saw_below_knee is technically accurate, but saw_non_above_knee_point might be more self-documenting. This is not a blocker — the current name is fine — but worth noting if you touch this code again.

jweber93:
Addressed in 7ff8783:

- Added the suggested comment at both `saw_below_knee = True` sites clarifying it marks any non-above-knee point regardless of read success.
- Added `test_recommendations_single_above_knee_point_message` and `test_recommendations_single_failed_point_no_tone_mapping_blame` covering the single-measurement above-knee and single-measurement failed-reading cases you flagged. Both pass.

Left the `saw_below_knee` → `saw_non_above_knee_point` rename as-is per your note (not blocking); happy to do it here or in a later pass if you'd rather.

---

PR #581, opened 2026-07-04
Fix report_payload averaging ΔE over all accumulated passes instead of latest pass

### LINKED ISSUE #577: Bug: report_payload averages ΔE over all accumulated passes, not the latest pass — headline pre/post ΔE, improvement %, and persisted history are diluted
**From the QE audit** (`QE_AUDIT_FINDINGS.md`, PR #576). **Severity: High · Confidence: High (reproduced).**

## Summary

`calibrator/reports.py:report_payload` computes each measurement bucket's `avg_de`/`max_de` (and `avg_gamma`) over the **raw, append-only** bucket list. Every other consumer of these buckets first reduces to the *latest pass* (grayscale) / *latest reading per item* (WB, CMS). When a bucket accumulates more than one pass — e.g. the operator re-measures a ramp and re-imports — the report blends stale earlier-pass readings into the headline numbers, and `record_session` then persists those wrong numbers into per-TV history, poisoning future LLM context.

## Evidence

`calibrator/reports.py:20-45`:

```python
def stats(measurements: list) -> Dict[str, Any]:
    ...
    items = [m_to_dict(m, target, signal_range, code_scale) for m in measurements]
    valid = [item["delta_e"] for item in items if item["delta_e"] is not None]
    return {"avg_de": round(sum(valid)/len(valid), 2) ..., "max_de": round(max(valid), 2) ...}

pre  = stats(session["pre_measurements"])   # <-- raw bucket, all passes
post = stats(session["post_measurements"])  # <-- raw bucket, all passes
wb   = stats(session["wb_measurements"])
cms  = stats(session["cms_measurements"])
gamma = gamma_stats(session["gamma_measurements"])
```

Every other reader reduces first:
- Live grayscale view: `latest_grayscale_pass(raw_meas, max_count=len(levels), …)` (`calibrator/session.py:1396`).
- Quality gate grayscale: `latest_grayscale_pass(...)` (`calibrator/quality.py:142`).
- Live gamma view: `latest_gamma_pass(...)` (`calibrator/session.py:1496`).
- Quality gate WB / CMS: `latest_wb_measurements(...)` / latest-reading-per-colour (`calibrator/quality.py:62,114-120`).

`latest_grayscale_pass` (`session.py:1177`) walks backwards and stops at a `ZRO_SESSION_BREAK_SECONDS` gap, so it returns only the most-recent ramp; `report_payload` never applies that reduction.

`improvement_pct` (`reports.py:46-48`) is derived from the diluted `pre`/`post`, so it is wrong too. `record_session` (`server.py:1669`) writes `pre_grayscale_avg_de` / `post_grayscale_avg_de` / `improvement_pct` from this report into `.calibration-history/…`.

## Reproduction

Seed `pre_measurements` with two timestamp-separated ramps — a poor first pass, then a good re-measured pass:

```
report_payload   pre_cal avg_de = 10.67   max_de = 23.29   (blends both passes)
latest_grayscale_pass truth      = 2.48    max_de = 3.98    (correct, latest ramp only)
```

## Failure Scenario

An operator measures a pre-cal ramp, adjusts, re-measures, and re-imports (two `POST /api/session/{sid}/import/zro` calls). Dedup is by timestamp+RGB, so the fresh re-measured ramp is appended, not merged. `pre_measurements` now holds two passes. The final report's headline "before" ΔE and the before/after improvement % are wrong, and the wrong values are written to per-TV history. The same happens for `post_measurements`, and analogously for gamma (`gamma_measurements`), WB (`wb_measurements`), and CMS (`cms_measurements`) during iterative tuning.

## Root Cause

`report_payload` is the only measurement consumer that treats an append-only bucket as a single coherent pass; it skips the per-bucket latest-pass / latest-per-item reduction the rest of the app applies.

## Fix

In `report_payload`, reduce each bucket before `stats()`, mirroring the existing helpers:
- [ ] grayscale `pre`/`post` → `latest_grayscale_pass(bucket, max_count=len(grayscale_levels_for_ramp(session["grayscale_ramp_steps"])), signal_range, code_scale)`
- [ ] gamma → `latest_gamma_pass(...)`
- [ ] WB → `latest_wb_measurements(...)`
- [ ] CMS → latest reading per colour (mirror `quality.py:114-120`)

## Test

`tests/test_delta_report.py::test_report_uses_latest_grayscale_pass`: seed `pre_measurements` with two timestamp-separated ramps (bad then good), assert `pre_cal.avg_de` equals the good-ramp average, not the blend. Add multi-pass CMS + gamma cases asserting report metrics match the quality gate's latest-only figures. Existing single-pass tests still pass (`latest_grayscale_pass` on one pass is identity).

## Affected File

- `calibrator/reports.py:20-48` (`report_payload` / `stats` / `gamma_stats`)

jweber93:
## 📺 Overview

Closes #577

### What does this PR do?
* **Component(s) affected:** `calibrator/reports.py` (report generation / persisted history)

---

* **The Problem:** `report_payload` computed each bucket's `avg_de`/`max_de` (and `avg_gamma`) over the raw, append-only measurement list. Every other consumer (live grayscale/gamma views, quality gates) first reduces to the latest pass / latest reading per item. When a bucket accumulates more than one pass (operator re-measures and re-imports), the headline pre/post ΔE, `improvement_pct`, and the values persisted to per-TV history via `record_session` were diluted by stale earlier-pass readings.
* **The Solution:** `report_payload` now reduces each bucket before computing stats, mirroring the existing helpers in `calibrator/session.py` / `calibrator/quality.py`:
  * grayscale `pre`/`post` → `latest_grayscale_pass(bucket, max_count=len(grayscale_levels_for_ramp(...)), signal_range, code_scale)`
  * gamma → `latest_gamma_pass(bucket, gamma_levels_for_session(session), signal_range, code_scale)`
  * WB → `latest_wb_measurements(bucket)` (latest gain/offset reading)
  * CMS → latest reading per colour (mirrors `quality.py`'s per-colour dedup)
* **Impact on existing workflows/APIs:** `report_payload`'s output shape is unchanged; only the underlying averages/measurement lists now reflect the latest pass. Also updated `calibrator/report_samples.py`'s fixture to include `timestamp`s (required for the new reduction) and aligned its sample gamma stimulus points to the tracked levels (20/40/60/80%).

---

### Verification Steps
1. `python -m pytest tests/test_delta_report.py tests/test_server_api.py tests/test_report_samples.py -q`
2. Full suite: `python -m pytest -q` — all passing.

Added `TestReportUsesLatestPassOnly` in `tests/test_delta_report.py` covering:
* multi-pass grayscale (bad pass + re-measured good pass) → report matches the good-pass-only average
* multi-reading CMS (stale + fresh reading for the same colour) → report uses the latest reading
* multi-pass gamma (stale + fresh reading for the same tracked level) → report uses the latest reading

Also fixed several pre-existing test fixtures in `tests/test_delta_report.py` / `tests/test_server_api.py` that lacked `timestamp`s (a prerequisite for the pass-reduction helpers) or used WB labels that didn't match the recognized gain/offset patterns.

---

## 🧼 Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_011Zjiett4BmjrMBxhFcQsVh)_

### DIFF
--- a/calibrator/report_samples.py
+++ b/calibrator/report_samples.py
@@ -21,6 +21,7 @@ def _measurement(
     y: float = 0.3290,
     x_scale: float = 0.95,
     z_scale: float = 1.09,
+    timestamp: str = "2026-04-19T09:30:00",
 ) -> Measurement:
     stimulus = round(stimulus_pct * 255 / 100)
     return Measurement(
@@ -31,6 +32,7 @@ def _measurement(
         Z=y_value * z_scale,
         label=label,
         stimulus_rgb=(stimulus, stimulus, stimulus),
+        timestamp=timestamp,
     )
 
 
@@ -58,10 +60,10 @@ def build_sample_report() -> Dict[str, Any]:
         _measurement("80% Gray", 80, 82.1, x=0.3128, y=0.3292),
     ]
     gamma_measurements = [
-        _measurement("10% Gray", 10, 2.6, x=0.3125, y=0.3289),
-        _measurement("25% Gray", 25, 9.7, x=0.3125, y=0.3290),
-        _measurement("50% Gray", 50, 27.4, x=0.3126, y=0.3290),
-        _measurement("75% Gray", 75, 58.9, x=0.3127, y=0.3291),
+        _measurement("20% Gray", 20, 9.7, x=0.3125, y=0.3289),
+        _measurement("40% Gray", 40, 27.4, x=0.3125, y=0.3290),
+        _measurement("60% Gray", 60, 58.9, x=0.3126, y=0.3290),
+        _measurement("80% Gray", 80, 89.3, x=0.3127, y=0.3291),
     ]
 
     session = {
--- a/calibrator/reports.py
+++ b/calibrator/reports.py
@@ -45,15 +45,30 @@ def report_payload(session: Dict[str, Any]) -> Dict[str, Any]:
             "measurements": items,
         }
 
+    # latest_grayscale_pass / latest_gamma_pass reduce by walking timestamps
+    # backwards; if a bucket carries no timestamps (e.g. legacy/pre-#577 data),
+    # that walk finds nothing to anchor on and returns empty. Fall back to
+    # treating the whole bucket as one coherent pass rather than reporting
+    # None/empty stats for measurements that are actually there.
     grayscale_levels = grayscale_levels_for_ramp(session.get("grayscale_ramp_steps", 11))
     latest_pre = latest_grayscale_pass(
         session["pre_measurements"], max_count=len(grayscale_levels),
         signal_range=signal_range, code_scale=code_scale,
     )
+    if not latest_pre and session["pre_measurements"]:
+        latest_pre = latest_grayscale_pass(
+            session["pre_measurements"], max_count=len(grayscale_levels),
+            signal_range=signal_range, code_scale=code_scale, detect_sessions=False,
+        )
     latest_post = latest_grayscale_pass(
         session["post_measurements"], max_count=len(grayscale_levels),
         signal_range=signal_range, code_scale=code_scale,
     )
+    if not latest_post and session["post_measurements"]:
+        latest_post = latest_grayscale_pass(
+            session["post_measurements"], max_count=len(grayscale_levels),
+            signal_range=signal_range, code_scale=code_scale, detect_sessions=False,
+        )
     latest_wb = latest_wb_measurements(session["wb_measurements"])
     latest_cms: Dict[str, Any] = {}
     for measurement in session["cms_measurements"]:
@@ -63,6 +78,11 @@ def report_payload(session: Dict[str, Any]) -> Dict[str, Any]:
     latest_gamma = latest_gamma_pass(
         session["gamma_measurements"], gamma_levels, signal_range, code_scale
     )
+    if not latest_gamma and session["gamma_measurements"]:
+        latest_gamma = latest_gamma_pass(
+            session["gamma_measurements"], gamma_levels, signal_range, code_scale,
+            detect_sessions=False,
+        )
 
     pre = stats(latest_pre)
     post = stats(latest_post)
--- a/calibrator/reports.py
+++ b/calibrator/reports.py
@@ -7,7 +7,14 @@ from typing import Any, Dict, List, Optional
 
 from fastapi import HTTPException
 
-from .session import m_to_dict
+from .session import (
+    gamma_levels_for_session,
+    grayscale_levels_for_ramp,
+    latest_gamma_pass,
+    latest_grayscale_pass,
+    latest_wb_measurements,
+    m_to_dict,
+)
 
 
 def report_payload(session: Dict[str, Any]) -> Dict[str, Any]:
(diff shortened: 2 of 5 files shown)


jweber93:
Actionable feedback (ordered by priority)

Guard / fallback for measurements without timestamps (recommended)
Problem: latest_grayscale_pass and latest_gamma_pass detect session breaks by timestamps and will return empty lists if no timestamps are present. That will change behavior for sessions with timestamp-less measurements (e.g., imported CSVs).
Fix (pick one):
Preferable: attempt timestamp-based extraction first, then fall back to treating the whole list as a single pass when the timestamp-based extraction returns empty but there are measurements. Example (reports.py): before or after current call to latest_grayscale_pass: if not latest_pre and session.get("pre_measurements"): latest_pre = latest_grayscale_pass( session["pre_measurements"], max_count=len(grayscale_levels), signal_range=signal_range, code_scale=code_scale, detect_sessions=False, )
Do the same fallback for post and gamma (use detect_sessions=False or a simple slice of the last N measurements).
Alternatively: detect whether measurements contain any timestamp and call latest_grayscale_pass(..., detect_sessions=False) when none do.
Add tests for the fallback case (imported CSV style): a test where measurements have no timestamps and confirm report_payload still produces reasonable stats (matching previous behavior).
Verify consumer expectations / API contract
Confirm downstream consumers of report_payload expect the new behavior (fewer samples, last-pass-only). If external code or persisted reports expect averages across the whole accumulation, you might need a migration note or an API version bump.
If you intend for this to be a breaking change, document it in CHANGELOG/PR description.
Tests / CI
You added focused unit tests — run the full test suite and CI to ensure nothing else relies on the old semantics (especially tests that implicitly relied on averaging all measurements).
Add tests that exercise:
Mixed: earlier (stale) pass followed by a later pass with timestamps (already covered).
No timestamps present (see item 1).
Interleaved timestamps with gaps that trigger session breaks at ZRO_SESSION_BREAK_SECONDS boundary.
Small code suggestions
In reports.py you still reference session[...] directly (same as before). If you prefer defensive coding, consider using session.get(...) with a sensible default for any places you expect missing keys.
Consider adding a short comment in reports.py explaining the intent: "reduce bucket to latest contiguous pass (by timestamp) before averaging — fallback for timestamp-less imports".
Ensure imports are minimal and ordered (not important for correctness but keeps lint happy).
Observability / debugging
When latest_* functions return empty (e.g., due to missing timestamps), it may be helpful to log/debug at info/trace level so operators can understand why a session reports empty stats. A single-line debug message in report_payload when a fallback is used is valuable.
Tests you added look correct — quick remarks
tests/test_delta_report.py: new TestReportUsesLatestPassOnly covers grayscale, CMS, gamma — good coverage.
tests updated to include timestamps — necessary for current implementation; reinforce item (1) by adding a test that simulates legacy/no-timestamp input.
Suggested quick patch (example fallback)

Add this immediately after you compute latest_pre / latest_post / latest_gamma in reports.py:

Fallback for imported/legacy measurements that lack timestamps:
if not latest_pre and session.get("pre_measurements"): latest_pre = latest_grayscale_pass( session["pre_measurements"], max_count=len(grayscale_levels), signal_range=signal_range, code_scale=code_scale, detect_sessions=False, )

Mirror for post and gamma (for gamma you might call latest_gamma_pass with the same levels but first try detect-by-timestamp then fallback to treating entire list).

What I did and what's next

I inspected the changed files in this PR (reports.py, session.py helpers, report_samples.py, and updated tests) and confirmed the new approach selects the latest contiguous pass via timestamp and then averages only that pass.
Next I recommend adding the timestamp-less fallback and new tests, then running CI. If you want, I can:
Draft the exact patch to add the fallback in reports.py and run a quick lexical search for other places that should be updated.
Or open specific review comments on the PR pointing to the lines to change.

jweber93:
Good catch on the timestamp-less fallback — pushed a fix:

- `report_payload` now falls back to `detect_sessions=False` (whole-bucket-as-one-pass) for grayscale pre/post and gamma whenever the timestamp-based reduction comes back empty but the bucket isn't. WB/CMS reductions don't depend on timestamps, so no fallback needed there.
- Extended `latest_gamma_pass` with the same `detect_sessions` flag `latest_grayscale_pass` already has, so the fallback has something to call.
- Added `test_report_falls_back_to_whole_bucket_when_no_timestamps` covering a legacy/no-timestamp session.

Confirmed production data always carries timestamps (both `zro_import.py` and `csv_adapter.py` stamp every measurement), so this fallback is purely a safety net for older persisted sessions or otherwise-unstamped input — not something exercised by the two real import paths today.

Left the other suggestions (API version bump, debug logging on fallback) out of scope for this fix — happy to follow up if you'd like those too. Full suite passes.

---

PR #582, opened 2026-07-04
Fix history_summary session_count saturating at display cap

### LINKED ISSUE #578: Bug: history_summary.session_count under-counts once a TV has more than 10 recorded sessions
**From the QE audit** (`QE_AUDIT_FINDINGS.md`, PR #576). **Severity: Low · Confidence: High.**

## Summary

`history_summary()` reports `session_count` using the display-capped loader, so once a TV has more than 10 recorded calibrations the count saturates at 10.

## Evidence

`calibrator/history.py:240-255`:

```python
def history_summary(tv_key: str) -> Dict[str, Any]:
    history = load_history(tv_key)          # <-- default limit=10
    ...
    return {
        "session_count": len(history),      # <-- capped at 10
        ...
    }
```

`load_history(tv_key, limit=10)` (`history.py:124`) truncates to `limit` before returning. The `latest_*` summary fields are still correct (they read index 0), only `session_count` is wrong. Exposed via `GET /api/session/{sid}/llm/history-summary` (`server.py:1382`).

## Failure Scenario

A display with 15 recorded calibrations: `history_summary()["session_count"]` returns `10`, understating the real history depth in the API/LLM-context field.

## Root Cause

The summary reuses the display-capped `load_history` instead of counting the full `sessions.jsonl`.

## Fix

Count lines directly (a small `count_sessions(tv_key)` that reads the file without the display cap), or pass a large/`None` limit for the count only.

## Test

`tests/test_history_metrics.py`: record 12 sessions, assert `history_summary()["session_count"] == 12`.

## Affected File

- `calibrator/history.py:245` (`history_summary`), `:124` (`load_history` default limit)

jweber93:
## 📺 Overview

Closes #578

### What does this PR do?
* **Component(s) affected:** `calibrator/history.py` (history summary / LLM context)

---

* **The Problem:** `history_summary()` computed `session_count` as `len(load_history(tv_key))`, but `load_history()` defaults to a display cap of `limit=10`. Once a TV had more than 10 recorded calibration sessions, `session_count` saturated at 10, understating the real history depth reported via `GET /api/session/{sid}/llm/history-summary`.
* **The Solution:** Added `count_sessions(tv_key)` which counts every valid JSON line in `sessions.jsonl` directly, without the display cap. `history_summary()` now uses `count_sessions()` for `session_count` instead of the length of the capped `load_history()` result. The `latest_*` fields are unaffected since they only ever read the first (most recent) entry.
* **Impact on existing workflows/APIs:** None beyond the corrected count — `load_history()`'s default cap and behavior are unchanged for callers that intentionally want the truncated, most-recent-first list (e.g. detailed history views).

---

### Verification Steps
1. Run `python -m pytest tests/test_history_metrics.py -q --no-cov`
2. New test `TestHistorySummarySessionCount::test_session_count_not_capped_at_display_limit` records 12 sessions and asserts `history_summary("u8g")["session_count"] == 12`.

---

## 🧼 Checklist
* [ ] Documentation (README, inline docstrings) updated to reflect changes. (not applicable — internal fix)

---
_Generated by [Claude Code](https://claude.ai/code/session_01NfPcmTiZU5xqVS2fPCvJ2t)_

### DIFF
--- a/calibrator/history.py
+++ b/calibrator/history.py
@@ -151,7 +151,10 @@ def count_sessions(tv_key: str) -> int:
     """Return the total number of recorded sessions for this TV.
 
     Unlike load_history(), this counts every line in sessions.jsonl and is
-    not capped by the display limit.
+    not capped by the display limit. A line that fails to parse as JSON
+    (e.g. a partial line from a concurrent in-progress append) is skipped
+    rather than counted, so this is a conservative lower bound if
+    record_session() is writing at the exact moment this is called.
     """
     path = _sessions_path(tv_key)
     if not path.exists():
@@ -268,6 +271,8 @@ def history_summary(tv_key: str) -> Dict[str, Any]:
     """Return a lightweight summary dict for API exposure.
 
     Keys: session_count, latest_date, latest_post_de, baseline_post_de.
+    session_count is the total number of recorded sessions (via
+    count_sessions), not capped by load_history()'s display limit.
     """
     history = load_history(tv_key)
     baseline = load_baseline(tv_key)
--- a/calibrator/history.py
+++ b/calibrator/history.py
@@ -147,6 +147,33 @@ def load_history(tv_key: str, limit: int = 10) -> List[Dict[str, Any]]:
     return list(reversed(entries))[:limit]
 
 
+def count_sessions(tv_key: str) -> int:
+    """Return the total number of recorded sessions for this TV.
+
+    Unlike load_history(), this counts every line in sessions.jsonl and is
+    not capped by the display limit.
+    """
+    path = _sessions_path(tv_key)
+    if not path.exists():
+        return 0
+
+    count = 0
+    try:
+        with open(path, encoding="utf-8") as fh:
+            for line in fh:
+                line = line.strip()
+                if line:
+                    try:
+                        json.loads(line)
+                        count += 1
+                    except json.JSONDecodeError:
+                        pass
+    except OSError:
+        return 0
+
+    return count
+
+
 def load_baseline(tv_key: str) -> Optional[Dict[str, Any]]:
     """Return the first-ever calibration baseline for this TV, or None."""
     path = _baseline_path(tv_key)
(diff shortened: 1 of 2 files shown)


jweber93:
Actionable suggestions (recommended changes / improvements)

Consider the partial-write / concurrent append case
Why: If another process is appending to sessions.jsonl while count_sessions reads it, you may encounter a partially-written last line which will be skipped by json.loads and thus undercount the true number of sessions.
Suggestion: either
live with the current behavior (it’s conservative — it skips bad lines), or
treat partially-parsed trailing data differently (e.g., retry reading the last partial line after a short sleep, or rely on writer-side atomic append semantics), or
add a comment in count_sessions docstring that partial writes may be skipped and that the writer should append atomically.
Action: add a small comment or a unit test that simulates an incomplete/partial line to document expected behavior.
Slight performance / reliability micro-optimizations
Observation: current implementation parses every line with json.loads to validate lines. If the file is large this is still O(n) and okay because it streams, but if you can rely on sessions.jsonl being one JSON object per non-empty line, counting non-empty lines is faster (no JSON parse).
Suggestion: if you control the writer and guarantee one JSON object per line, replace the json.loads check with a line-count (sum(1 for line in fh if line.strip())) and optionally provide a validate=True parameter that parses lines when callers want validation.
Action: either keep the current defensive behavior or add a short docstring note about the trade-off and/or add an optional parameter to count_sessions.
Add a test for malformed/partial lines
Why: The new test asserts a straightforward happy path where 12 well-formed sessions are recorded. Add a test that simulates a partially-written or malformed line in sessions.jsonl to assert the function’s behavior is as expected (skip or raise).
Suggestion: write a small test that opens the sessions file and appends a truncated JSON line, then assert count_sessions returns the expected conservative count (or whatever behavior you choose).
Action: add the test as a follow-up commit.
Small nits / style

If calibrator/history.py already imports json and typing names used here, nothing more to change; if not, ensure json is imported and typing annotations used consistently.
Consider adding a one-line docstring to history_summary() or to the exported API docs to mention that session_count is the total recorded (not the display cap).
Non-blocking possible refactor

If more session-file statistics are needed later (e.g., earliest/latest timestamps, incomplete-line detection), consider a small sessions-file utility module that exposes count/iterate/validate functions and centralizes concurrency/format rules.

jweber93:
Thanks for the review — addressed the actionable items in 8853f9b:

- Documented in `count_sessions()` and `history_summary()` docstrings that an unparseable trailing line (e.g. a concurrent in-progress append) is conservatively skipped rather than counted.
- Added `test_partial_trailing_line_is_skipped_conservatively`, which appends a truncated JSON line after 3 recorded sessions and asserts `count_sessions()` still returns 3 without crashing.

Left the `json.loads`-vs-line-count micro-optimization and the sessions-file utility module refactor as-is since you flagged both as optional/non-blocking trade-offs — happy to revisit if you'd like either done now.

---

PR #611, opened 2026-07-10
fix: correct gamma tooltip target from 2.40 to 2.2 (#608)

### LINKED ISSUE #608: Low (frontend): gamma column tooltip says "BT.1886 targets ≈ 2.40" while every gate, chart, and target in the app uses 2.2
**From the QE audit** (`QE_AUDIT.md` bug-hunting pass, 2026-07-10). Severity **Low** (operator-facing guidance text, no computation affected), Confidence **High**.

## Defect

`frontend/src/components/MeasurementTable.jsx:14`:

```js
const GAMMA_COL = { header: 'γ', key: 'effective_gamma',
  tooltip: 'Effective gamma — the measured power-law exponent at each stimulus step. BT.1886 targets ≈ 2.40.' };
```

Everything else in the product targets **2.2** for SDR:

- `SDR_TARGET.gamma = 2.2` (`calcore/models.py:318-324`) — drives `m_to_dict()` reference luminance and `effective_gamma` shown in this very column.
- `GammaChart.jsx` / `OverlayGammaChart.jsx` draw the reference line at `2.2` for non-PQ.
- `calcore/targets.py:50-51` uses gamma 2.2 for SDR BT.1886 targets ("per SDR_TARGET in models.py").
- Quality gate `QG_GAMMA_THRESHOLD` compares against `target.gamma` (2.2) with ±0.05 tolerance (`calibrator/quality.py:86-109`).
- QE_AUDIT ground truth: "SDR = BT.1886 EOTF / gamma 2.2".

## Failure scenario

An operator on the Gamma or Post-Grayscale step reads the tooltip, sees their measured γ ≈ 2.2 with the tooltip claiming the target is ≈ 2.40, and manually darkens midtones toward 2.4 — which then **fails** the app's own gamma quality gate (|γ − 2.2| ≤ 0.05) and the chart's 2.2 reference line, sending them in circles between the tooltip and the gate.

## Fix

Change the tooltip to state the app's actual target, e.g.: `"Effective gamma — the measured power-law exponent at each stimulus step. SDR target: 2.2 (PQ/HDR tracks 1.0)."` Ideally derive the number from the session target/EOTF like `GammaChart` does rather than hardcoding.

**Model tier:** 1 — local/low-cost model fine.

## Regression test

Vitest/RTL on `MeasurementTable` with `includeGamma`: tooltip text asserts the 2.2 SDR target (guards against the copy drifting from `SDR_TARGET.gamma` again).

jweber93:
## 📺 Overview

Fixes incorrect gamma column tooltip that stated BT.1886 targets ≈ 2.40, while every gate, chart, and target in the app uses 2.2.

Closes #608

### What does this PR do?

-   **Component(s) affected:** frontend/src/components/MeasurementTable.jsx, frontend/src/components/MeasurementTable.test.jsx

---

-   **The Problem:** The gamma column tooltip in MeasurementTable said "BT.1886 targets ≈ 2.40" while the app's actual SDR gamma target is 2.2 (SDR_TARGET.gamma, GammaChart reference line, quality gate threshold). An operator could be misled into darkening midtones toward 2.4, which would then fail the app's own quality gate (|γ − 2.2| ≤ 0.05).
-   **The Solution:** Updated tooltip text to "SDR target: 2.2 (PQ/HDR tracks 1.0)" to match the app's actual targets. Added Vitest/RTL test suite for MeasurementTable including a regression test that verifies the tooltip contains "2.2" and not "2.40".
-   **Impact on existing workflows/APIs:** None. Purely a UI text correction. No computation or data flow affected.

---

### Environment Details

-   **OS / Target Display:** macOS (CI)
-   **Hardware/Mock Used:** Unit tests only (no hardware required)

### Verification Steps

1.  Run  — no errors
2.  Run  — all 141 tests pass including 5 new MeasurementTable tests
3.  Run  — static/ rebuilt successfully
4.  E2E tests: pre-existing failures confirmed on main (not introduced by this change)

### Firmware / TV Settings

-   Not applicable — UI text change only

---

-   Tooltip text corrected from "BT.1886 targets ≈ 2.40" to "SDR target: 2.2 (PQ/HDR tracks 1.0)"

---

## 🧼 Checklist

-   [x] Code adheres to project style guidelines (formatting, typing, linting).
-   [x] Unit tests added/updated and passing.
-   [x] Documentation (README, inline docstrings) updated to reflect changes.
-   [x] No credentials, local absolute paths, or hardware-specific hardcoding leaked.


jweber93:
Possible Improvements
1. Dynamically source the target value:
Instead of hardcoding "2.2" in the tooltip text, derive it from the session target or SDR_TARGET.gamma constant (as suggested in issue #608). This would ensure tooltip and gate never drift apart again without a code change.

2. Tooltip accessibility enhancement:
The tooltip interaction uses hover + tooltip trigger selection. Consider adding keyboard navigation support (e.g., focus-visible state) to make the tooltip more discoverable for keyboard-only users.

jweber93:
Verify secondary fix context:

Confirm why "spotread" was flagged (bundler size/hygiene check docs or error message).
Ensure "Argyll default" is accurate terminology (not misleading vs. actual Argyll behavior).

jweber93:
**Verify secondary fix context:**

`spotread` was flagged by `tests/test_static_ui.py:test_bundle_no_argyllcms_references` (line 81), which asserts that no ArgyllCMS binary/tool names appear in the frontend JS bundle. The test's intent is to catch accidental backend imports or hardcoded paths leaking into client JS — not just for size, but to prevent internal tool names from appearing in production-facing code.

"Argyll default" is accurate: the integration-mode label describes the default behavior of spotread, which is an ArgyllCMS tool. The original label named the specific binary (`spotread`) rather than the toolkit (`ArgyllCMS`). Since the test flags any ArgyllCMS binary name, replacing `spotread` with `Argyll` (the toolkit name used by the parent component and the `ArgyllCard` component itself) is the correct fix while keeping the label meaningful to operators familiar with ArgyllCMS.

jweber93:
**Item 1 — Dynamically sourced gamma target:** Addressed in 5e7a9f7. Added `SDR_GAMMA_TARGET = 2.2` as a named export in `frontend/src/utils/fmt.js` (adjacent to the existing `isPqEotf` function that already documents the 2.2 vs 1.0 distinction). The `MeasurementTable` tooltip now uses a template literal referencing this constant, so tooltip and test stay in sync with the source of truth. The regression test also imports and asserts against `SDR_GAMMA_TARGET` directly, so drifting the constant will break the test.

**Item 2 — Tooltip keyboard accessibility:** The `Tooltip` component already has `tabIndex={0}`, `onFocus`/`onBlur` handlers, and `role="button"` — keyboard users can tab to the ⓘ trigger and focus will show the tooltip. The existing implementation covers this use case.

Issue #603, opened 2026-07-10
Critical: analyze() color-patch targets use a fixed 100-nit scale — false color ΔE on any display not at exactly 100 nits


jweber93:
**From the QE audit** (`QE_AUDIT.md` bug-hunting pass, 2026-07-10). Severity **Critical**, Confidence **High** (traced + reproduced).

## Defect

`analyze()` compares **absolute measured XYZ (nits)** against color-patch targets that live on a **fixed 0–100 scale**, while grayscale targets are correctly scaled by the measured display peak. Any display whose peak is not exactly 100 nits reports large color ΔE even when perfectly calibrated.

## Failure scenario (reproduced)

Simulate a *perfectly calibrated* 300-nit BT.709 SDR display (measured XYZ generated from the exact target math, gamma 2.2) and run `analyze()`:

```
grayscale_avg_de = 0.000          # correct — grayscale targets scale by measured peak
color 255,0,0  bucket=100  dE2000 = 25.06   target Y=21.3, measured Y=63.8
color 0,255,0  bucket=100  dE2000 = 25.04
color 0,0,255  bucket=100  dE2000 = 20.86
```

A flawless display is reported as catastrophically wrong on every color patch. These `color_100_*` / `color_75_*` metrics feed `determine_phase()` (CMS unlock / verify gating, `calcore/phase.py:37-44`), the LLM adjustment prompts (`calcore/llm.py`), and `assess_convergence()` — so the pipeline can never "converge" on colors and the LLM is asked to fix nonexistent errors.

<details><summary>Repro script</summary>

```python
from calcore.analysis import analyze
from calcore.models import AnalysisConfig, Patch
from calcore.spaces import BT709_RGB_TO_XYZ

PEAK, GAMMA = 300.0, 2.2
def ideal(r, g, b, code_max=255):
    lin = [(c / code_max) ** GAMMA for c in (r, g, b)]
    return tuple(sum(BT709_RGB_TO_XYZ[row][i] * lin[i] for i in range(3)) * PEAK for row in range(3))

patches = [Patch(f"{c},{c},{c}", c, c, c, ideal(c, c, c), kind="grayscale") for c in (0, 64, 128, 191, 255)]
patches += [Patch("255,0,0", 255, 0, 0, ideal(255, 0, 0), kind="color"),
            Patch("0,255,0", 0, 255, 0, ideal(0, 255, 0), kind="color"),
[...]

PR #612, opened 2026-07-10
Fix #604: linearize color-patch targets with the session EOTF

### LINKED ISSUE #604: Critical: color-patch targets skip EOTF linearization — 75%-stimulus color patches compared against a wrong reference (ΔE ~6.7 on a perfect display)
**From the QE audit** (`QE_AUDIT.md` bug-hunting pass, 2026-07-10). Severity **Critical**, Confidence **High** (traced + reproduced). Companion to #603 (same code path, independent root cause — fixing #603 alone does not fix this).

## Defect

`target_xyz_for_patch()` converts **gamma-encoded signal values** straight through the linear RGB→XYZ matrix for color patches. Grayscale targets correctly apply the session EOTF (PQ / BT.1886 / gamma) to derive target luminance; color targets never do. For any non-100% color patch the target luminance is inflated by the missing EOTF (at 75% signal, gamma 2.2: target Y is ~41% too bright).

## Failure scenario (reproduced)

Perfectly calibrated 100-nit BT.709 display (so #603's peak-scale term is exactly zero), 75% red patch `(191,0,0)`:

```
grayscale_avg_de = 0.000
color 255,0,0  bucket=100  dE2000 = 0.00    # 100% patch: signal == linear, unaffected
color 191,0,0  bucket= 75  dE2000 = 6.67    # 75% patch: target not linearized
```

The **entire `color_75_*` metric family is wrong by construction** — the "75" saturation bucket exists precisely for these patches (`calcore/models.py:45-58`), and `determine_phase()` gates CMS→verify on `color_75_avg_de <= 3.0` (`calcore/phase.py:40-44`). A perfect display fails that gate (6.67 > 3.0), and the LLM is prompted to "fix" 75%-patch errors that don't exist (the QE-audit target of dE2000 < 2.0 is unreachable).

<details><summary>Repro script</summary>

```python
from calcore.analysis import analyze
from calcore.models import AnalysisConfig, Patch
from calcore.spaces import BT709_RGB_TO_XYZ

PEAK, GAMMA = 100.0, 2.2
def ideal(r, g, b, code_max=255):
    lin = [(c / code_max) ** GAMMA for c in (r, g, b)]
    return tuple(sum(BT709_RGB_TO_XYZ[row][i] * lin[i] for i in range(3)) * PEAK for row in range(3))

patches = [Patch(f"{c},{c},{c}", c, c, c, ideal(c, c, c), kind="grayscale") for c in (0, 128, 255)]
patches += [Patch("255,0,0", 255, 0, 0, ideal(255, 0, 0), kind="color"),
            Patch("191,0,0", 191, 0, 0, ideal(191, 0, 0), kind="color")]
s = analyze(patches, AnalysisConfig(mode="sdr", eotf="gamma22", target_space="bt709", code_max=255))
for row in s.color_rows: print(row["label"], row["bucket"], round(row["dE2000"], 2))
# 255,0,0 100 0.0
# 191,0,0 75  6.67
```
</details>

## Evidence / code path

- `calcore/targets.py:36-40` — `target_rgb` is normalized **signal** (code / code_max), no EOTF applied.
- `calcore/targets.py:62-63` — for `not patch.is_grayscale`, that signal vector goes directly into `rgb_to_xyz()` (a linear-light matrix, `calcore/spaces.py:42-50`).
- Contrast with the grayscale branch (`calcore/targets.py:42-60`), which runs the same normalized code through `pq_target_nits` / `bt1886_eotf` / `gamma_eotf` before building XYZ.
- Reached from every color analysis: `analyze()` (`calcore/analysis.py:129`) ← `_run_llm_background` / `/gamut/diagnosis` / `/suggested-patches` / `/next-settings` (`server.py:1131, 1632, 2101, 2382`).
- `tests/test_calcore/test_calcore.py:83-98` encodes the buggy behavior: its "near-target" 75% red measured XYZ (30.93, 15.95, 1.45) equals the *non-linearized* signal target, i.e. a physically implausible measurement for a gamma-2.2 display.

## Root cause

The color-target path treats gamma-encoded signal RGB as linear light; the reference for a color patch must be `EOTF(signal)` per channel (the same session EOTF the grayscale branch already selects), then matrixed to XYZ.

## Fix

In `target_xyz_for_patch()`, linearize each channel of `target_rgb` with the session EOTF (share the mode/eotf dispatch already used by the grayscale branch) before `rgb_to_xyz()`, then apply the peak scaling from #603. Update the crafted measurements in `test_calcore.py` to physically-correct values.

**Model tier:** 4-5 (EOTF + color-space math, invalidates existing golden expectations) — frontier model recommended.

## Regression test

`tests/test_calcore/`: `test_analyze_75pct_color_de_is_zero_for_perfect_display` — 75% R/G/B patches with EOTF-correct measured XYZ must yield `color_75_avg_de < 0.5` for eotf ∈ {gamma22, bt1886, pq}.

jweber93:
## 📺 Overview

Closes #604

### What does this PR do?
* **Component(s) affected:** calcore (color-patch target computation / analysis pipeline)

---

* **The Problem:** `target_xyz_for_patch()` fed gamma/PQ/BT.1886-encoded signal RGB straight into the linear-light RGB→XYZ matrix for color patches, while grayscale targets correctly ran the same signal through the session EOTF first. A 100% patch was unaffected (signal == linear at the top of the range), but any non-100% color patch — most notably the "75%" saturation bucket that gates CMS→verify phase progression — got a badly inflated target luminance (ΔE ~6.7 on a perfectly calibrated display, per the issue repro).
* **The Solution:** Factored the EOTF dispatch (PQ / BT.1886 / gamma) that grayscale targets already used into a shared helper, `_linear_fraction_of_peak()`. Color-patch channels are now linearized with it before being matrixed to XYZ, then the existing #603 peak-scaling is applied on top. The grayscale branch was refactored to call the same helper (verified mathematically equivalent to the prior grayscale-only code) so both patch kinds share one EOTF dispatch, per the issue's suggested fix.
* **Impact on existing workflows/APIs:** No signature changes. Only the numeric value of color-patch targets changes (grayscale targets are unaffected). This corrects `color_75_avg_de` / `color_100_avg_de` reporting and the CMS→verify phase gate in `calcore/phase.py`, which reads those values directly.

* [x] Matrix transformations or LUT calculations have been validated.
* [x] Floating-point precision or data truncation risks have been addressed.

---

### Verification Steps
1. `python -m pytest tests/` — full suite passes (382 tests, 82.6% coverage).
2. Ran the issue's repro script directly against the fixed code: 75% red patch `(191,0,0)` on a perfect 100-nit BT.709 display now reports `dE2000 = 0.0` (previously `6.67`).
3. Updated the "near-target" fixture in `tests/test_calcore/test_calcore.py` (previously encoded the buggy non-linearized target as the measurement) to physically-correct EOTF-linearized values.
4. Added `test_analyze_75pct_color_de_is_zero_for_perfect_display`, covering 75% R/G/B patches for all three supported EOTFs (gamma22, bt1886, pq), asserting `color_75_avg_de < 0.5` on a perfect display.

---

## 🧼 Checklist
* [ ] Documentation (README, inline docstrings) updated to reflect changes. — not applicable, no public API/behavioral surface beyond the bug fix itself.

### DIFF
--- a/calcore/targets.py
+++ b/calcore/targets.py
@@ -22,20 +22,34 @@ def _linear_fraction_of_peak(
     Shares the session EOTF dispatch (PQ / BT.1886 / gamma) so grayscale and
     color-patch targets linearize signal the same way before either is
     turned into an XYZ target (see issue #604 — color patches previously
-    skipped this step entirely).
+    skipped this step entirely). The return value is always a fraction in
+    [0, 1] where 1.0 represents *peak*; multiply by an absolute peak (nits)
+    to get an absolute target luminance, or feed several channels' fractions
+    directly into rgb_to_xyz() for a color-patch target.
+
+    Per-EOTF calling convention:
+    - PQ: pq_target_nits() returns absolute ST.2084 nits for signal *n*,
+      clipped at *peak*; dividing by *peak* expresses that as the same
+      0..1 fraction the other branches return.
+    - BT.1886: bt1886_eotf(v, lw, lb, gamma) expects white/black levels
+      (lw/lb) in the same units. Passing lw=1.0 and lb=black_frac (the
+      display's measured black normalized to *peak* = 1.0) returns the
+      fractional curve directly — algebraically identical to calling
+      bt1886_eotf(n, peak, measured_black_y, gamma) and then dividing by
+      peak, just without the extra division.
+    - gamma: n**gamma is already a 0..1 fraction of peak (n=1 -> 1.0), no
+      rescaling needed.
     """
-    if cfg.mode.lower() == "hdr" or cfg.eotf.lower() == "pq":
+    eotf = (cfg.eotf or "gamma22").lower()
+    mode = (cfg.mode or "sdr").lower()
+    if mode == "hdr" or eotf == "pq":
         return pq_target_nits(n, peak) / peak
-    elif cfg.eotf.lower() == "bt1886":
-        gamma = 2.2 if cfg.mode.lower() == "sdr" else 2.4
+    elif eotf == "bt1886":
+        gamma = 2.2 if mode == "sdr" else 2.4
         black_frac = measured_black_y / peak if peak > 0 else 0.0
         return bt1886_eotf(n, 1.0, black_frac, gamma=gamma)
     else:
-        gamma = (
-            2.2
-            if cfg.eotf.lower() in ("gamma22", "2.2", "gamma")
-            else float(cfg.eotf)
-        )
+        gamma = 2.2 if eotf in ("gamma22", "2.2", "gamma") else float(eotf)
         return gamma_eotf(n, gamma=gamma)
 
 
--- a/calcore/targets.py
+++ b/calcore/targets.py
@@ -11,6 +11,34 @@ from .spaces import detect_matrix, rgb_to_xyz
 logger = logging.getLogger(__name__)
 
 
+def _linear_fraction_of_peak(
+    n: float,
+    cfg: AnalysisConfig,
+    peak: float,
+    measured_black_y: float,
+) -> float:
+    """Linearize a normalized code value *n* into a 0..1 fraction of peak luminance.
+
+    Shares the session EOTF dispatch (PQ / BT.1886 / gamma) so grayscale and
+    color-patch targets linearize signal the same way before either is
+    turned into an XYZ target (see issue #604 — color patches previously
+    skipped this step entirely).
+    """
+    if cfg.mode.lower() == "hdr" or cfg.eotf.lower() == "pq":
+        return pq_target_nits(n, peak) / peak
+    elif cfg.eotf.lower() == "bt1886":
+        gamma = 2.2 if cfg.mode.lower() == "sdr" else 2.4
+        black_frac = measured_black_y / peak if peak > 0 else 0.0
+        return bt1886_eotf(n, 1.0, black_frac, gamma=gamma)
+    else:
+        gamma = (
+            2.2
+            if cfg.eotf.lower() in ("gamma22", "2.2", "gamma")
+            else float(cfg.eotf)
+        )
+        return gamma_eotf(n, gamma=gamma)
+
+
 def target_xyz_for_patch(
     patch: Patch,
     code_max: int,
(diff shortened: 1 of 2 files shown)


jweber93:
Concrete items to address before merging (actionable)

Defensive handling: guard against missing/None cfg.eotf or cfg.mode

Problem: _linear_fraction_of_peak calls cfg.mode.lower() and cfg.eotf.lower() unguarded. If an AnalysisConfig instance ever has eotf or mode set to None or an unexpected type, this will raise AttributeError.
Suggested fix: normalize config values earlier or defensively coerce to str: e.g.
eotf = (cfg.eotf or "gamma22").lower()
mode = (cfg.mode or "sdr").lower()
Use those local variables in the function.
Why: minimal risk and cheap to add; protects from malformed configs or future code paths.
Zero / tiny peak handling and bt1886 arguments

Problem: The bt1886 branch computes black_frac = measured_black_y / peak if peak > 0 else 0.0 and calls bt1886_eotf(n, 1.0, black_frac, ...). Confirm bt1886_eotf accepts (n, peak, black_y, gamma) where black_y is absolute (so passing black_frac with peak=1.0 is correct). If bt1886_eotf expects absolute black Y (not fraction), this would be a bug.
Suggested action: add a short comment documenting the bt1886_eotf calling convention (we want to return a fraction, so calling with peak=1.0 and black_frac is intentional). Optionally add an assertion/test that bt1886_eotf(..., peak=1.0, black_y=black_frac) == bt1886_eotf(..., peak=peak, black_y=measured_black_y) / peak.
Why: removes possible confusion for future readers and avoids subtle numeric mistakes.
Possible improvements (non-blocking, actionable)

Add a small unit test that covers measured_peak_y == None or 0.0 and measured_black_y edge cases:
Verify the fallback peak == 100.0 behavior produces stable results and does not cause ZeroDivisionError.
Add docstring to _linear_fraction_of_peak describing:
return value is a 0..1 fraction of peak,
calling conventions for PQ / BT.1886 / gamma branches (especially the bt1886 parameterization),
explicit note why grayscale and color-patch linearization must match.
Add a short regression test that exercises the CMS→verify gate (the phase that uses color_75_avg_de) to ensure the behavioral fix prevents false gating. The existing test proves dE for 75% patches is small; a direct test of the phase gate logic will prevent future regressions.
Optional micro-optimization: cache cfg.mode.lower() and cfg.eotf.lower() into local variables in the helper to avoid repeated .lower() calls.
Other small nits

The comment "peak = measured_peak_y if measured_peak_y and measured_peak_y > 0 else 100.0" appears twice (previously one). Consider a small helper or a single normalized peak variable earlier in the file to avoid duplication and centralize the default 100.0 choice.
If you want maximum clarity, include a one-line note in the PR description about the bt1886 argument transformation (peak -> 1.0 + black_frac) so reviewers know it's intentional.

jweber93:
Addressed in 804089f:

- `_linear_fraction_of_peak()` now coerces `cfg.eotf`/`cfg.mode` defensively (`(cfg.eotf or "gamma22").lower()`) instead of assuming non-`None`, cached as locals (also covers the micro-optimization).
- Expanded the docstring to document the per-EOTF calling convention, including why the BT.1886 branch calls `bt1886_eotf(n, 1.0, black_frac, gamma)` (fractional form) rather than the absolute `(peak, measured_black_y)` form the grayscale caller used to use — confirmed algebraically equivalent, now written down.
- Added `test_target_xyz_for_patch_falls_back_to_100_peak_when_measured_peak_missing` (measured_peak_y `None`/`0.0`, confirms no `ZeroDivisionError` and correct fallback).
- Added `test_linear_fraction_of_peak_guards_against_none_mode_and_eotf`.
- Added `test_color_75_fix_unlocks_cms_to_verify_phase_gate`, an end-to-end test through `analyze()` → `determine_phase()` proving a perfect display's 75% patches now actually unlock CMS→verify (they were stuck failing the `<= 3.0` gate before this PR).

On the "appears twice" nit: checked the current diff — `peak = measured_peak_y if measured_peak_y and measured_peak_y > 0 else 100.0` now appears exactly once in `targets.py` (used by both branches), so I left it as is rather than introduce a cross-file helper with `analysis.py`'s differently-defaulted (mode-dependent) fallback.

Full suite + ruff still green.

---

